PR #65, opened 2026-02-11
Fixed issue with incorrect URL within a message

Ricisss:
**Problem**
If the user sends a message with text `[http://localhost:5173](http://localhost:5173)` and the front end tries to make a URL object from it, it throws an exception that results in a complete black screen.

**Fix**
Checking with URL.canParse before that solves this issue.

**Note**
I think the root of the problem is even before the message gets to database, because in database the message looks like this:
`<p><strong>[</strong><a target="_blank" rel="noopener noreferrer nofollow" href="http://localhost:5173](http://localhost:5173)"><strong>http://localhost:5173](http://localhost:5173)</strong></a></p>`

Which, to me, looks very incorrect, but I have failed to find the actual cause of this weird parsing. I suspect tiptap, but I have no experience with this library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,8 @@ Once the server is running, open your web browser and navigate to [http://localh
 2. Open Dev Tools (`CTRL + Shift + I` or `Right Click > Inspect`)
 3. Open the console
 4. Type useToken('your_token_here')
+> [!NOTE]
+> You have to wrap your token in single quotes ' or double quotes ".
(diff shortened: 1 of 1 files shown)


diogomartino:
Yea the database message looks really strange, might need extra investigation later. The pr looks good but I'll add an extra try catch  around the whole serializer just to make sure a problem here just doesn't crash the whole interface.

Issue #48, opened 2026-02-10
Add abilitiy to remove user entirely as owner of server


schemen:
Currently, I don't seem to be able to remove a user from the server, only kick them from an active session.

I'd like to be able to delete a user entirely.

christian45410:
++ This - I have a few unused account I'd like to remove as its bloating up the user mgmt area.

P00DLEMAN:
Same. Would like to remove "Sharkord" user, if possible.

PR #82, opened 2026-02-11
feat: add user deletion functionality in admin panel

Vigno04:
- Add delete-user endpoint with proper permissions and validation
- Add USER_DELETED activity log type
- Add delete user option in admin users table dropdown menu
- Prevent self-deletion and show confirmation dialog
- Update user list after successful deletion
- Fix React controlled input warning in text input dialog

Backend:
- New delete-user route with MANAGE_USERS permission check
- Disconnect user if online before deletion
- Log deletion activity with reason

Frontend:
- Delete user option in table dropdown (hidden for own user)
- Confirmation dialog with optional reason input
- Auto-refresh user list after deletion
- Fix text input dialog controlled component warning


schemen:
This would resolve #48

apps/client/src/components/server-screens/server-settings/users/table-user.tsx
```diff
@@ -3,30 +3,65 @@ import {
...
   const roles = useUserRoles(user.id);
   const status = useUserStatus(user.id);
+  const ownUserId = useOwnUserId();
 
   const onModerateClick = useCallback(() => {
     setModViewOpen(true, user.id);
   }, [user.id]);
 
+  const onDeleteUser = useCallback(async () => {
+    const reason = await requestTextInput({
```
diogomartino: No need to ask a reason. When you disconnect the user just send a generic reason like "Your account has been deleted" or something

apps/server/src/routers/users/delete-user.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    // Close the websocket connection if the user is connected
+    const userWs = ctx.getUserWs(input.userId);
+
+    if (userWs) {
+      userWs.close(DisconnectCode.KICKED, input.reason ?? 'User deleted');
+    }
+
+    // Delete the user from the database
+    await db.delete(users).where(eq(users.id, input.userId));
```
diogomartino, line 246: You should check if the user exists before trying to delete it.

apps/server/src/routers/users/delete-user.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    // Publish user deletion event
+    publishUser(input.userId, 'delete');
```
diogomartino, line 249: The delete event is not doing anything in the client, it's missing all that implementation to cleanup all things related to that users for all connected users.

apps/client/src/components/server-screens/server-settings/users/table-user.tsx
```diff
@@ -3,30 +3,49 @@ import {
...
   const status = useUserStatus(user.id);
+  const ownUserId = useOwnUserId();
 
   const onModerateClick = useCallback(() => {
     setModViewOpen(true, user.id);
   }, [user.id]);
 
+  const onDeleteUser = useCallback(async () => {
+    const trpc = getTRPCClient();
+
```
diogomartino, line 72: I asked you to remove the asking for a reason but it should still ask for confirmation xD

 Check `requestConfirmation` method.

apps/client/src/features/server/slice.ts
```diff
@@ -256,6 +256,21 @@ export const serverSlice = createSlice({
...
+      for (const channelId in state.typingMap) {
+        state.typingMap[channelId] = state.typingMap[channelId].filter(
+          (id) => id !== userId
+        );
+      }
+
+      for (const channelId in state.voiceMap) {
+        delete state.voiceMap[channelId].users[userId];
+      }
+    },
```
diogomartino, line 273: This is only deleting the users but it still needs to delete everything that is linked to that user like messages, possible emojis created by the user, etc.

Vigno04: For the messages isn't it better to have something like reddit where if the user is deleted then instead of the username it tells you deleted? You tell ne

diogomartino: I also thought about it but I honestly don't know what's better tbh.

stb109: I personally think the messages should stay and the username just set to deleted or something, or have the option of deleting the messages as well.

Vigno04: Yeah that seems the best option, choice is always the best, tomorrow I will implement that

diogomartino: Alright, in that case you need to change your approach. The user needs to stay in the database but add a new column called `deleted` with default false. When you delete you only change that flag and you set the username as "Deleted". The UI also needs to show this, maybe a strike through or something.

Vigno04: Will look into that

Aasikki: Yes, being able to choose would be best! It would allow to keep messages when deleting an account that isn't used anymore, but it's also important to have the option to delete the messages when deleting an account because of spam/harassing or other bad behavior.

Vigno04: Now it should be resolved: when you try to delete a user, it asks if you want to delete emoji files and messages. If you say no, they get assigned to a deleted user. This way, if you want to delete them later, you can, and also there is no modification to the database. The only "problem" is that in the UI, they get grouped, so if two users write a message one after another and you delete both, then it appears like both are sent by a deleted user—but I think it's okay for the extra benefits we get from this implementation.

diogomartino: Thanks, I'll check it. This one is pretty big and tricky, I'll look at it when I have some free time. Thank you :heart_hands:

diogomartino:
It's not perfect but pretty good. I'll make some adjustments after I merge. Ty!

Issue #86, opened 2026-02-11
XSS Issue


kruitbosdotdev:
I checked the application quickly and found out that websockets send HTML:

```
{
  "id": 48,
  "method": "mutation",
  "params": {
    "input": {
      "content": "<p>I wonder if i can type in other channels</p>",  // Right here
      "channelId": 5,
      "files": []
    },
    "path": "messages.send"
  }
}
```

By using a tool like BurpSuite we can customize websocket messages and modify the HTML:

```
{
  "id": 48,
  "method": "mutation",
  "params": {
    "input": {
      "content": "<h1 style=\"font-size: 100px; background: linear-gradient(90deg, red, orange, yellow, green, blue, indigo, violet); -webkit-background-clip: text; -webkit-text-fill-color: transparent;\"I found a way to do some xss :)</h1>",  // Right here
      "channelId": 5,
      "files": []
    },
    "path": "messages.send"
  }
}
```

This results in the following:

<img width="845" height="400" alt="Image" src="https://github.com/user-attachments/assets/dae5265d-41fc-4e36-955a-6149198ae57f" />

Frontend validation works as expected, but the frontend should **never** be in control. Sanitize the messages on the backend before storing it.

PR #91, opened 2026-02-11
fix(#86): prevent xss

diogomartino:
#86


RossBrunton:
Allowing "class" still allows attackers to use arbitrary css classes, right?

Why does `class` need to be an allowed attribute?

diogomartino:
I was thinking about possible plugins that might want to set specific classes on their messages. Idk if it makes sense, but I kept it. But classes are not really dangerous as scripts are so that's the most important part. I might revisit that in the future and make it bullet proof.

PR #100, opened 2026-02-11
WebRTC Server for Port Simplification. 

cgingue11:
Swapping from WebRTC port per connection to a server to broker. Simplifies port mappings to only require one open port for WebRTC.

Also fixed ICE announcedAddress not being configurable and thus broken behind reverse proxies not running on the same host as Sharkord.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,6 @@ Upon first run, Sharkord will generate a default configuration file located at `
 | `port` | `4991` | The port number on which the server will listen for HTTP and WebSocket connections |
 | `debug` | `false` | Enable debug logging for detailed server logs and diagnostics |
 | `autoupdate` | `false` | When enabled, it will automatically check for and install updates with no user intervention |
-| `maxFiles` | `40` | Maximum number of files that can be uploaded in a single request |
-| `maxFileSize` | `100` | Maximum file size in megabytes (MB) allowed per uploaded file |
 | `webrtcPort` | `40000` | The port for the WebRTC server to listen on (UDP+TCP) |
 | `announcedAddress` | _(empty)_ | When set, announces this address for WebRTC connections. Useful when your reverse proxy is on a different host than Sharkord. |
 


jrhodnik:
Nice work with the `announcedAddress`. Only suggestion is to add a blurb to the README to let people behind reverse proxies know how to configure it.

cgingue11:
> Nice work with the `announcedAddress`. Only suggestion is to add a blurb to the README to let people behind reverse proxies know how to configure it.

Added some info to the README!

I don't use Caddy or NGINX. I'd really appreciate if someone with working configs for either could share their set up for a more accurate config.

diogomartino:
Damn, didn't really know that was possible. If that really works and only a single port is required, it's huge. Gonna test it out and investigate a little bit more before merging.

jrhodnik:
@cgingue11 I was able to get this running via a reverse proxy, Caddy specifically, however Sharkord was actually auto-detecting my public IP properly so I didn't need this patch. 

I detailed the issues we had, including the resolution, in #114 . I'll hammer a bit more this evening to see if I can get Sharkord to run in debug.

it's 100% a problem on my end and TS/JS isn't my primary language so I just gotta struggle through it 😅

EDIT: I just now noticed your docs about using raw sockets! I'll try to give it that a shot here this evening as well!

cgingue11:
> @cgingue11 I was able to get this running via a reverse proxy, Caddy specifically, however Sharkord was actually auto-detecting my public IP properly so I didn't need this patch.
> 
> I detailed the issues we had, including the resolution, in #114 . I'll hammer a bit more this evening to see if I can get Sharkord to run in debug.
> 
> it's 100% a problem on my end and TS/JS isn't my primary language so I just gotta struggle through it 😅
> 
> EDIT: I just now noticed your docs about using raw sockets! I'll try to give it that a shot here this evening as well!

So, you got the single webrtc port working for you? That's excellent news!

jlroskens:
There's also #157 which addresses this. I think I prefer how the rtcMinPort and rtcMaxPort were handled in that they weren't necessarily deprecated but were available as a way to toggle between these functionalities. I'm also running this on a k8s cluster and had to make these same changes prior to seeing there's already 2 PRs for it.

This one seems like it's got the most attention though. Wondering if maybe the other PR's approach to handling these ports can be incorporated? If not, maybe just remove them entirely instead of deprecating them since this is alpha. Just my opinion, for what its worth.

diogomartino:
@jlroskens yea I agree, no deprecations at this stage. If we gonna change that, we completely remove the other options. I'll need some time to dig into this and the other one before merging.

cgingue11:
I'm not too proud to admit when somebody else does it better - i do think the other PR is more elegant than mine. I do however agree that removing min max port is the way to go. If we get this webrtc multiplexing working, I see no value in being able to open 20 ports instead of one, so I'd like to see some convergence of our two methods. I trust your judgement @diogomartino and of course i have no real say in how you reconcile the differences between our PRs. Just excited to see this project getting so much love!

diogomartino:
@cgingue11 I'll merge this one first and get things from #157. That one is more complex and not working locally, but it has some good ideas there. I just need you to:

- resolve conflicts
- remove `rtcMinPort` and `rtcMaxPort` for good
- add the port config in `webrtc.port` so it's less confusing.

In the #157 he used the 7882 port because he states it's the standard port for WebRTC but I would probably still keep 40000 as the default, because if we change that now when people upgrade it's gonna break because 7882 will be closed.

cgingue11:
@diogomartino Resolved conflicts, removed references to min/max port. 

Hope this is sorted now!

diogomartino (approved):
LGTM

PR #106, opened 2026-02-11
Feature: Floating Media Controls

AbyssalSoda:
Implemented floating media control buttons and corresponding hooks to better match standard video chat applications.

- Controls Bar component renders a standard pill overlay for mic mute/unmute | webcam on/off | screenshare start/stop
- Standardized buttons via ControlToggleButton component for icon swapping & tool tips
- useControlsBarVisibility for auto fade/slide away effects

No current documentation provided for contributions; so changes made based on the practices demonstrated within the code base itself.

### DIFF
--- a/apps/client/src/components/channel-view/voice/control-toggle-button.tsx
+++ b/apps/client/src/components/channel-view/voice/control-toggle-button.tsx
@@ -39,7 +39,7 @@ const ControlToggleButton = memo(
           variant="ghost"
           size="icon"
           className={cn(
-            'rounded-[16px] h-11 w-11 transition-all duration-200',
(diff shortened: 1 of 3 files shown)


AbyssalSoda:
> I think it would be better if it only showed when the mouse is really close to the bottom, I'd say 80 or 90%

Gotcha, set it to 80%
#
> the design looks cool but it's a bit off with the rest of the app, specially the border radius

Removed some of the hard coded values here, altered a explicit return, and converted the 'pill' into a brick by reducing bevels from rounded-xl to rounded-md to match Sharkords native bevels. I also separated the disconnect button and increased its width 18px to give it some visual prominence (closer to discord).

Pill design will be missed o7

PR #109, opened 2026-02-11
feat: docker-compose and env var reading

deranjer:
Hey, wanted to add an example docker-compose for those of us that like to go that route. I really prefer env vars for configuring my service when using docker-compose, but it seems like that wasn't an option, so I added that as well.  This did bump dependencies since I needed to add types/node dep which bumped other things.

For the env vars I did require they be prepended with SHARKORD, which differs from the .ini file, but I don't want it to read in something as esoteric as `port` since that could be sitting in env somewhere if someone runs it outside of docker.

I steered away from adding in anything related to caddy or nginx for this since I know that can vary a lot, and I personally have caddy configured via labels. I can look at adding that if it is needed.


TheBlckbird:
What about setting the prefix to just `SK_`? I think it makes the config a bit nicer.

apps/server/src/config.ts
```diff
@@ -5,6 +5,18 @@ import { getPrivateIp, getPublicIp } from './helpers/network';
...
+  interface Env {
+    // SHARKORD_ prefixed environment variables
+    SHARKORD_PORT?: string;
+    SHARKORD_DEBUG?: string;
+    SHARKORD_MAX_FILES?: string;
+    SHARKORD_MAX_FILE_SIZE?: string;
+    SHARKORD_RTC_MIN_PORT?: string;
+    SHARKORD_RTC_MAX_PORT?: string;
+  }
+}
```
diogomartino: Please move this to `apps/server/src/declarations.d.ts` and remove `SHARKORD_MAX_FILES` and `SHARKORD_MAX_FILE_SIZE` since they were a leftover and are not used for anything.

apps/server/src/config.ts
```diff
@@ -57,10 +69,34 @@ const text = await fs.readFile(CONFIG_INI_PATH, {
...
+  config.server.debug = process.env.SHARKORD_DEBUG === 'true' || process.env.SHARKORD_DEBUG === '1';
+}
+
+if (process.env.SHARKORD_MAX_FILES) {
+  config.http.maxFiles = parseInt(process.env.SHARKORD_MAX_FILES, 10);
+}
+
+if (process.env.SHARKORD_MAX_FILE_SIZE) {
+  config.http.maxFileSize = parseInt(process.env.SHARKORD_MAX_FILE_SIZE, 10);
+}
```
diogomartino: Ignore these ones

docs/docker/docker-compose.yaml
```diff
@@ -0,0 +1,83 @@
...
+    # Can override config file settings
+    environment:
+
+      # Server configuration
+      SHARKORD_PORT: "4991"
+      SHARKORD_DEBUG: "false"
+      
+      # HTTP configuration
+      SHARKORD_MAX_FILES: "40"
+      SHARKORD_MAX_FILE_SIZE: "100" # MB
```
diogomartino: Remove

docs/docker/docker-compose.yaml
```diff
@@ -0,0 +1,83 @@
...
+#   Enable verbose logging (default: false)
+#
+# SHARKORD_MAX_FILE_SIZE
+#   Maximum file upload size in MB (default: 100)
+#
+# SHARKORD_RTC_MIN_PORT / SHARKORD_RTC_MAX_PORT
+#   UDP port range for WebRTC media (default: 40000-40020)
+#   Increase range if you need more concurrent voice connections
+#
+
```
diogomartino: Maybe move this to the main readme file instead of having it here

diogomartino (changes requested):

deranjer:
Okay, these should all be resolved, I also pulled the port options listed in the readme.

@TheBlckbird I'm open to changing the ENV var names, but I do prefer to be as explicit as possible with those to avoid any name collision for people running this outside of docker, although I personally can't imagine going back to that.  If someone views their env vars I want them to immediately know what it is for.

diogomartino (approved):
LGTM

PR #116, opened 2026-02-12
Fixed: Voice Control Indication

AbyssalSoda:
**Issue**: If a user who does not have a webcam attempts to press the camera button, the button will turn green to indicate the webcam is ON despite the message "_Requested device not found_" being rendered. A similar issue arises  when pressing the button to share screens, as if the user decides to cancel they will prompted with a "_Permission denied by user_" message, yet the screen share icon will still remain blue and the screen itself will split to indicate sharing content.

**Fix:** Added rollback and fail closed cleanup for the webcam/screenshare controls on TRPC mutation errors.

### DIFF
--- a/apps/client/src/components/voice-provider/hooks/use-voice-controls.ts
+++ b/apps/client/src/components/voice-provider/hooks/use-voice-controls.ts
@@ -110,9 +110,15 @@ const useVoiceControls = ({
       await trpc.voice.updateState.mutate({
         webcamEnabled: newState
       });
-    } catch (error) {
+     } catch (error) {
       updateOwnVoiceState({ webcamEnabled: false });
-      await trpc.voice.updateState.mutate({ webcamEnabled: false }).catch(() => {});
+
+      try {
+        await trpc.voice.updateState.mutate({ webcamEnabled: false });
+      } catch {
+        // ignore
+      }
+
       toast.error(getTrpcError(error, 'Failed to update webcam state'));
     }
   }, [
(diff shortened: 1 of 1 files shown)


apps/client/src/components/voice-provider/hooks/use-voice-controls.ts
```diff
@@ -100,16 +101,18 @@ const useVoiceControls = ({
...
       } else {
         stopWebcamStream();
       }
+
+      await trpc.voice.updateState.mutate({
+        webcamEnabled: newState
+      });
     } catch (error) {
+      updateOwnVoiceState({ webcamEnabled: false });
+      await trpc.voice.updateState.mutate({ webcamEnabled: false }).catch(() => {});
```
diogomartino: Don't do this, do:

```
try {
 // code
} catch {
 // ignore
}
```

to keep the code consistent

apps/client/src/components/voice-provider/hooks/use-voice-controls.ts
```diff
@@ -132,32 +135,35 @@ const useVoiceControls = ({
...
+        
         // handle native screen share end
         video.onended = async () => {
           stopScreenShareStream();
           updateOwnVoiceState({ sharingScreen: false });
 
           await trpc.voice.updateState.mutate({
             sharingScreen: false
-          });
+          }).catch(() => {});
```
diogomartino: Same as above

apps/client/src/components/voice-provider/hooks/use-voice-controls.ts
AbyssalSoda: Gotcha, I've gone ahead and implemented the try/catch formatting based on your example.

Issue #98, opened 2026-02-11
Request: Add paper clip file upload option


dillfrescott:
Right now I believe you can only upload images from either dragging them or copy pasting them. I didn't see a way to open a file picker.

toby-short:
Done, made a pull request for it (PR #118) 😊

jreoka:
Cool thank you so much!

PR #118, opened 2026-02-12
Feature (#98) - Add paperclip file upload option.

toby-short:
Added a simple button to the input bar on text channel view for uploading files, created a handler for opening a browse window and linked it back to the upload file functionality.

#98

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,2 @@
+nix_direnv_manual_reload
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /cenas.md
 /DockerCenas
 /experiments
+.direnv/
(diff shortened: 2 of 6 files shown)


apps/client/src/hooks/use-upload-files.ts
```diff
@@ -142,7 +209,16 @@ const useUploadFiles = (disabled: boolean = false) => {
...
   }, [addFiles, can, settings, disabled]);
 
-  return { files, removeFile, filesRef, clearFiles, uploading, uploadingSize };
+  const fileInputProps = {
+    ref: fileInputRef,
+    type: 'file' as const,
+    multiple: true,
+    onChange: onFileDialogChange,
+    style: { display: 'none' },
+  };
```
diogomartino: Memoize with `useMemo` here.

toby-short: Will sort this now :)

PR #122, opened 2026-02-12
feat: add multi-language support with English and Italian translations

Vigno04:
- Implement i18n using react-i18next
- Add language provider and locale formatting hooks
- Update all UI components to use translations
- Add Italian translations (manually reviewed)
- English translations based on existing strings
- Key extraction and organization assisted by AI tools
- Code reviewed and tested by human
- Added explenation on how to add new languages


Ricisss:
If this gets merged, I could do Czech translations.

Vigno04:
If you want it's pretty easy to do, if you go on my branch you still see all the information, to start you just copy the english file and translate all the string

diogomartino:
I understand people wanting this but I don't really think it's a good time to add this, I would prefer to add it when the app is more stable.

Vigno04:
My thought about adding this now is so that in the future there is no need to go and find all the hard coded text in the app, this way from now on its standardized with a pretty good library so it's already set up, it support the fallback to English so even if a translation is half done there won't be problem in the visualization

diogomartino:
I understand, but adding this now will raise the amount of things I need to worry about. Not a priority at the moment, gonna close and implement in the future. Sorry about any wasted work you got here :/

Issue #125, opened 2026-02-12
Prevent Users from sending empty messages


GoldcrafterXD:
Users can send empty messages. Maybe there should be a check if a Message has Text or an Attachment before allowing it to be sent.

ghost:
Can confirm.

The check should also disregard spaces.

PR #127, opened 2026-02-12
Prevent Users from sending empty messages #125

GoldcrafterXD:
Users can no longer send/edit to empty messages, or messages with only spaces

### DIFF
--- a/apps/server/src/helpers/check-empty-message.ts
+++ b/apps/server/src/helpers/check-empty-message.ts
@@ -1,9 +1,0 @@
-const checkEmptyMessage = (content: string): boolean => {
-
-    // Matches strings that are either empty, contain only whitespace, or consist solely of empty or only space filled <p> tags (which currently is always the case for these message)
-    const checkEmptyRegex = /^(?:\s*|<p>\s*<\/p>)$/;
-
-    return checkEmptyRegex.test(content); 
-}
-
-export { checkEmptyMessage };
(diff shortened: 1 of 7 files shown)


apps/server/src/helpers/check-empty-message.ts
diogomartino, line 1: Checking for html tags is probably not the best idea here. Use this instead:

```
const isEmptyMessage = (content: string): boolean => {
  if (!content) return true;

  const text = content
    .replace(/<[^>]+>/g, '')     // remove all HTML tags
    .replace(/&nbsp;/g, ' ')     // normalize nbsp
    .trim();

  return text.length === 0;
};
```

Also try to keep the naming consistent with the rest of the code so in this case it would be `isEmptyMessage`.

apps/server/src/routers/messages/edit-message.ts
```diff
@@ -48,8 +49,16 @@ const editMessageRoute = protectedProcedure
       }
     );
 
+    if (checkEmptyMessage(input.content)) {
+      throw new Error('Message cannot be empty.');
```
diogomartino: Do not throw messages like this in the route handlers, use the `invariant` util. Check the other routes to see how it's used.

diogomartino (changes requested):

diogomartino:
Also it might be a good idea to check if the message is empty before sending it to the server.

GoldcrafterXD:
Changed

- Renamed checkEmptyMessage to isEmptyMessage and is now in sharkord/shared
- Improved isEmptyMessage
- Now using invariant in server
- Added check if Message is empty to Client

PR #148, opened 2026-02-13
Feature (#4) - Added screenshare audio capability as well as improved the audio stream control.

toby-short:
I've added the ability to stream screenshare audio whereas previously this was just video. It's set up to use a newly created StreamKind (SCREEN_AUDIO) keeping it separate from the mic input. 

As they were kept separate, I also added the ability to adjust screenshare audio volume and made changes to the volume controller to show screenshare audio sources and have their volume also adjustable.

image

All is working as intended, only issue with this is due to it being a browser restriction and the app not having any native connectivity, when streaming a window or whole screen, other users will be able to hear themselves. this is just because in the browser we don't have access to the native audio hooks for each app to be able to exclude it from the source.

Linked back to feature request #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@
   [![Mediasoup](https://img.shields.io/badge/Mediasoup-v3.19.11-green.svg)](https://mediasoup.org)
 </div>
 
+[![ko-fi](https://ko-fi.com/img/githubbutton_sm.svg)](https://ko-fi.com/B0B71U3476)
+
 ## What is Sharkord?
 
 > [!NOTE]
--- a/apps/client/package.json
+++ b/apps/client/package.json
@@ -1,7 +1,7 @@
 {
   "name": "client",
   "private": true,
-  "version": "0.0.1",
+  "version": "0.0.3",
   "type": "module",
   "scripts": {
     "dev": "vite",
--- a/apps/client/src/components/channel-view/voice/control-toggle-button.tsx
+++ b/apps/client/src/components/channel-view/voice/control-toggle-button.tsx
@@ -1,0 +1,60 @@
+import { Button } from '@/components/ui/button';
+import { Tooltip } from '@/components/ui/tooltip';
+import { cn } from '@/lib/utils';
+import { memo } from 'react';
+
+type TIconComponent = React.ComponentType<{ size?: number; className?: string }>;
+
+type TControlToggleButtonProps = {
+  enabled: boolean;
+  enabledLabel: string;
+  disabledLabel: string;
+  enabledIcon: TIconComponent;
+  disabledIcon: TIconComponent;
+
+  enabledClassName: string;
+  disabledClassName?: string;
+
+  onClick: () => void;
+  disabled?: boolean;
+};
+
+const ControlToggleButton = memo(
+  ({
+    enabled,
+    enabledLabel,
+    disabledLabel,
+    enabledIcon: EnabledIcon,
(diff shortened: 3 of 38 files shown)


cosvak:
Question, is there a way to point sharkord to listen to specific devices on the system instead of just grabbing the system default device. This could be a potential work around to stop chat echoing till native audio hooks can be enabled?

toby-short:
I've been doing some digging into audio devices, and yeah we can direct different audio to specific audio devices, but depending on the windows audio setup it may still pick it up, will keep looking into it

stickman89:
I haven't been able to make this feature work as intended. The screenshare sound is very quiet, almost inaudible, and strangely linked to the user's volume control and not the screenshare volume control for the user. At first we thought the game sound was coming back through the mic - due to how quiet it was, so we tested it with a user who muted their mic... Game sound continued to play faintly whilst he was muted, but the screenshare volume control had no effect. I found I was only able to adjust the game volume by adjusting the user volume slider, but doing that affected the user volume and screenshare stream at the same time, but the game volume remained almost inaudible. We tested 'entire screen' and the individual application window... The same issue occurred for both.

I've just confirmed, as I noticed you were using Chrome... This feature is only working as intended in Chrome. It causes the above issues in Firefox.

cosvak:
> I've been doing some digging into audio devices, and yeah we can direct different audio to specific audio devices, but depending on the windows audio setup it may still pick it up, will keep looking into it

Thank you so much, I use a workaround that includes creating a virtual audio device in windows using [VB AUDIO CABLE](https://vb-audio.com/Cable/) then in the control panel > Manage Audio Devices > RECORDING > I activate the LISTEN TO THIS DEVICE option on VB CABLE to hear everything running to it. Then I use [EARTRUMPET](https://eartrumpet.app/) to forward my desired app I want to push the audio of to the chat while still hearing everything.

I use this kind of setup back in the day on discord to run OBS streams in screensharing while passing the audio through to a non conflicting device.

I think this kind of setup will also help bypass the issues @stickman89 is experiencing on firefox. Which i can also confirm I also have experience the same issues they are as well. Using chromium based browsers seems to be key at the moment. I would suggest [Helium](https://helium.computer/) it's an open-sourced version of chromium with an emphasis on privacy and all the extra google bs stripped out.

Giggitybyte:
> I've just confirmed, as I noticed you were using Chrome... This feature is only working as intended in Chrome. It causes the above issues in Firefox.

I am also experiencing a similar problem using Firefox on Kubuntu. I tested with a friend and he was unable to hear any audio from what I was screensharing, but rather he heard my microphone through my screenshare audio stream. So, whenever I spoke, he heard me twice. He could mute either my voice channel audio stream or my screnshare audio stream, but he'd still be able to hear my voice.

The majority of my friend group is on some flavor of Linux and most everybody in the group (regardless of OS) uses Firefox. Asking everybody to install and use Chrome or any fork of Chrome is simply to use Sharkord is simply not an option. 

@toby-short I do hope you can figure out the issue. If there's any logs or browser data I can provide to help you debug, please let me know.

Rand-o:
audio share on linux doesnt seem to work, can only share the audio of the browser that sharkord is open in. Couldn't tell if that is a limitation of linux or not

v0idMrK:
Same here. On Fedora Linux with KDE screenshare sound does not seem to work (receiving or sending).

Maybe I am doing something wrong but I don't know what.

Thank you for this project :) It's great.

Issue #104, opened 2026-02-11
Input field and messages doesnt wrap if no spaces in between


Katacc:
Input field doesnt wrap and just continues outside of the borders

Katacc:
Additional information: It properly wraps if yo uuse real words, but if you spam letters, it doesnt wrap them

Katacc:
More information: It kindof breaks the whole site layout.. Don't send long strings without spaces

toby-short:
Got a fix for this about to pull request

PR #169, opened 2026-02-14
Bug (#104) - Fixed text channel overflowing.

toby-short:
Fixed text channel inputs overflowing, now wraps even without spaces, had to add some min-width to the main content area and the input wrapper, plus overflow and max-height on the input content itself.

Also fixes the messages themselves overflowing under the same conditions.

Relates to bug #104


apps/client/src/components/tiptap-input/index.tsx
```diff
@@ -184,10 +184,11 @@ const TiptapInput = memo(
     }, [editor, disabled]);
 
     return (
-      <div className="flex flex-1 items-center gap-2">
+      <div className="flex flex-1 items-center gap-2 min-w-0">
         <EditorContent
           editor={editor}
-          className={`border p-2 rounded w-full min-h-[40px] tiptap ${
+          style={{ maxHeight: '5rem' }}
```
diogomartino: add this as a class to keep things consistent

`max-h-[5rem]`

toby-short: Done 🙂

diogomartino (approved):
LGTM

Issue #170, opened 2026-02-14
Custom emote counts as empty message. Still present in 0.0.4


evoandroidevo:
Trying to send just a custom emote will give you a popup for can't send empty messages after updating to 0.0.3

toby-short:
Fixed this, included in PR #173 for next release.

evoandroidevo:
Still present in 0.0.4. Sending only a custom emotes wont send doesnt give a message now for why. trying to edit a message to only a custom emote says the message cant be empty.

diogomartino:
Should be fixed now

PR #173, opened 2026-02-14
Bug (#170) - Fixed sending messages with only an emote in.

toby-short:
Fixed sending messages with only emotes, changed the regex to just detect empty p tags or ones with line breaks/br tags in them.

Bug ref: #170

### DIFF
--- a/packages/shared/src/helpers/is-empty-message.ts
+++ b/packages/shared/src/helpers/is-empty-message.ts
@@ -3,7 +3,7 @@ const isEmptyMessage = (content: string): boolean => {
   if (!content) return true;
 
   const text = content
-    .replace(/<[^>]+>/g, '')
+    .replace(/<p\b[^>]*>(?:\s|&nbsp;|<br\b[^>]*>)*<\/p>/gi, '')
     .replace(/&nbsp;/g, ' ')
     .trim();
 


packages/shared/src/helpers/is-empty-message.ts
```diff
@@ -3,7 +3,7 @@ const isEmptyMessage = (content: string): boolean => {
   if (!content) return true;
 
   const text = content
-    .replace(/<[^>]+>/g, '')
+    .replace(/<p\b[^>]*>(?:\s|&nbsp;|<br\b[^>]*>)*<\/p>/gi, '')
```
diogomartino, line 6: I didn't think about the emojis when I made this, but I think your solution will open things too much. Forging the request will bypass this easily I think. Try this:

```typescript
const isEmptyMessage = (content: string): boolean => {
  if (!content) return true;

  // check if it has media (eg: emojis will be detected here)
  const hasMedia = /<(img|video|audio|iframe)\b/i.test(content);

  const cleaned = content
    // remove PM separators
    .replace(/<img[^>]*ProseMirror-separator[^>]*>/gi, "")
    .replace(/<br[^>]*ProseMirror-trailingBreak[^>]*>/gi, "")
    // remove all remaining tags
    .replace(/<[^>]*>/g, "")
    // normalize spaces
    .replace(/&nbsp;/gi, " ")
    .replace(/\u00A0/g, " ")
    .trim();

  const hasText = cleaned.length > 0;

  return !hasText && !hasMedia;
};

export { isEmptyMessage };
```

diogomartino:
Gonna merge it and make the change myself because I'll release the new version soon and I want this to be included, ty!

evoandroidevo:
#170 still present in 0.0.4, i have added a comment to the issue

PR #177, opened 2026-02-14
feat(128): add toggle to hide non-video participants in voice channels

### LINKED ISSUE #128: Allow us to "hide non-video participants"
Another feature towards Discord parity. Having the user icons taking up the same amount of space as streams makes viewing multiple streams difficult.

Image

caellach:
Implement option to selectively display only participants with active video streams in voice channels. The feature intelligently shows all participants when no video streams are active, preventing an empty channel view.

Changes:
- Add hideNonVideoParticipants state to Redux store with localStorage persistence
- Create VoiceOptionsController component in top bar with settings dropdown
- Implement smart filtering logic in voice channel view:
  - Only hides non-video users when video streams are present
  - Screen shares always visible (always include video)
  - External streams filtered based on video track presence
- Add selector, action creator, and hook following existing patterns

Resolves #128

### DIFF
--- a/apps/client/src/components/top-bar/voice-options-controller.tsx
+++ b/apps/client/src/components/top-bar/voice-options-controller.tsx
@@ -33,19 +33,21 @@ const VoiceOptionsController = memo(() => {
       </PopoverTrigger>
       <PopoverContent align="end" className="w-64">
         <div className="space-y-3">
-          <h4 className="font-medium text-sm mb-3">Voice Options</h4>
+          <h4 className="font-medium text-sm cursor-default mb-3">Voice Options</h4>
 
           <div className="flex items-center justify-between space-x-3">
-            <label
-              htmlFor="hide-non-video"
+            <span
+              onClick={() => handleToggleHideNonVideo(!hideNonVideoParticipants)}
               className="text-sm text-foreground cursor-pointer select-none flex-1"
             >
               Hide non-video participants
-            </label>
+            </span>
             <Switch
               id="hide-non-video"
               checked={hideNonVideoParticipants}
               onCheckedChange={handleToggleHideNonVideo}
+              data-1p-ignore
+              data-lpignore="true"
             />
           </div>
         </div>


diogomartino (changes requested):
Couple of things:
- no need to have a separate button on top for volume and for this. Join both volume controls and the hide non video in a single popover (keep the cog icon)
- for me it doesn't make sense to make that setting persistent, it should always start as false anytime the user joins a voice channel

zdlovett:
> for me it doesn't make sense to make that setting persistent, it should always start as false anytime the user joins a voice channel

I might not be understanding this point, but as a user I would always prefer a grid view, and not showing the non-streaming user-squares in the chat by default. The sidebar is more than good enough to know whos in the call, so if I'm looking at the window its just to see the streams. 

(also huge :heart: for this project, it amazing to have a self-hosted 3rd space)

diogomartino:
@zdlovett as a user I would expect that any time I join a voice channel I see every user that's in it, I wouldn't like to set that option today, forget about it, join tomorrow and be confused why the users are not showing. How do other apps work?

Giggitybyte:
> a user I would expect that any time I join a voice channel I see every user that's in it ...

I feel that the voice channel itself accomplishes this goal; it displays all users currently in the voice channel right under its name. Seeing users in the center pane doesn't seem all that important to me; as I usually switch to a text channel if nobody is actively streaming, both in Discord and Sharkord. Only if somebody is actively streaming will I pull up the voice channel in the center pane.

> How do other apps work?

For Discord, if nobody is actively screensharing or has a webcam enabled, it'll simply tell you that you have 'Show Non-Video Participants' disabled:

Screenshot_20260216_161957

Screenshot_20260216_161826

I do think that such a setting is important to have. When multiple people are in a voice channel and only two or three of them are screensharing, it makes it much easier to see each stream at once when the non-video participants of the channel are hidden.

The screenshots in #128 articulate this pretty well.

zdlovett:
I think this is easier to explain with some screenshots. 

When hanging out with people if I have the voice channel selected the listing under the channel name "Channel Alpha" in this case, is really all thats needed in order to tell who's in the call. Meanwhile the rest of the window is most useful as the streams from the other folks that are also hanging out. This is the Discord settings "grid view" and with "show non-video participants" toggled off. 

image

99% of the time when we are hanging out these are the settings that I would want, so having a way to default to this instead of messing with the setting every time would be a meaningful upgrade over Discord to me. 

At the moment in Sharkord the best we can do is this:

image

Which ends up with 50% of the space totally wasted with redundant user icons. I know that you can "pin" a stream and it will go to full size, but the idea here is to be able to see everyone's stream. 

Hopefully that helps to clarify what I'm trying to convey.

diogomartino:
@zdlovett alright, you convinced me. But I would say it's missing some visual queue saying that's enabled or something. Or are most of the users already used just to enable that and keep it that way?

zdlovett:
>  Or are most of the users already used just to enable that and keep it that way?

Certainly everyone in my friend group is used to it working this way, and one of main things we complain about in Discord is that it doesn't remember this toggle, so you need to set the grid view + don't show participants every time. 

I think that if nobody is streaming having some text explaining the existence of the relevant controls like what is shown in @Giggitybyte 's screenshot is a good way to handle discoverability.

diogomartino:
Agreed. @caellach please add that text and also create or link a ticket to this pr (see https://github.com/Sharkord/sharkord/blob/development/CONTRIBUTING.md)

diogomartino (approved):

diogomartino:
Solve the conflicts so I can merge, it should mostly be imports. I moved a lot of code, sorry about that.

caellach:
> Agreed. @caellach please add that text and also create or link a ticket to this pr (see https://github.com/Sharkord/sharkord/blob/development/CONTRIBUTING.md)

@diogomartino Relating to the text, would you prefer to show that instead of just showing the VoiceUserCards? Currently if there are no video streams of any kind then it reverts to displaying those, as if the option were disabled.

diogomartino:
That's also fine, but I'm thinking I would also add a permanent small visual queue to signal that the option is enabled, I just don't know a good place for it. Maybe an icon in the top bar next to the buttons that toggles off the option once clicked? Idk.

caellach:
> That's also fine, but I'm thinking I would also add a permanent small visual queue to signal that the option is enabled, I just don't know a good place for it. Maybe an icon in the top bar next to the buttons that toggles off the option once clicked? Idk.

I'd lean towards not adding a toolbar indicator for this. The revert-to-showing-everyone behavior when there's no video already handles that case well, so there's no real confusion there. The quick-disable shortcut is a nice thought, but I'm not sure it justifies a permanent toolbar slot for a single toggle. The toggle is one click away in the popover, and stream availability per user is already visible in the Channels View. If someone wonders why a participant isn't showing up, that's already answered without any new UI. It also sets a precedent we can't really follow through on: eventually most toggles will live in a popover or submenu, and we can't add a toolbar icon for each one. Even a generic "something is active in here" dot on the popover button breaks down once you have two toggles, since one of them will almost always be non-default.

I'd be curious what specific confusion you're anticipating though, since that might change my thinking.

diogomartino:
I'm not really a discord user so I'm not used to those features so maybe that's why I'm thinking this way. But if I joined a voice channel, toggled to hide, disconnected and then the next day joined the channel again (with users in it already) and saw no one I would be confused for a while before I figure it out. But it's just me probably, I'll merge :+1:

diogomartino (approved):
LGTM

PR #179, opened 2026-02-14
Resolve #155: User Popovers in voice and channel names in context menu

### LINKED ISSUE #155: Right clicking a user in a voice channel opens the context menu for the channel
It would be more intuitive if it opened a context menu for the user instead

also maybe show the users profile when clicking on them

Herbycheck:
As mentioned in #155, right clicking on a user connected to a voice channel opens the context menu for the voice channel, not the user. 
I added a label to the context menu with the channels name so its clear that the admin is editing/deleting the channel and not the user. Also, when clicking on a user in the voice channels' user list it now opens a popover, same as in the member list.

I only used existing components, and I'm relatively new to React. Please tell me if I did something wrong.
Resolves #155


diogomartino (approved):
LGTM

diogomartino:
No, you did a good job, it's an improvement for sure, ty

Issue #178, opened 2026-02-14
Link processing SSRF


ihopeitdoes:
Fairly self explanatory, simply by pasting a link into the chat we can make requests to internal ip address ranges of the server. Server blindly makes requests to any https url passed as a message. It also incorrectly extracts urls fom messages.
You'll get things like this extracted:
`'https://localhost:8000/sample.jpg">https://localhost:8000/sample.jpg</a></p>',` instead of just `https://localhost:8000/sample.jpg
`
To mitigate SSRF see here https://github.com/Sharkord/sharkord/blob/development/apps/server/src/queues/message-metadata/get-message-metadata.ts#L28 and here: https://github.com/OP-Engineering/link-preview-js?tab=readme-ov-file#ssrf-concerns

PR #181, opened 2026-02-14
fix(#178): prevent SSRF

diogomartino:
#178

Also improves URL detection inside messages.


ihopeitdoes:
Thanks for the quick fix

Issue #180, opened 2026-02-14
Different channel name character limits when creating/editing


Herbycheck:
When creating a channel the character limit is 16 characters, when editing a channel the length is 24 characters.

<img width="525" height="396" alt="Character Limit when creating a channel is 16 characters" src="https://github.com/user-attachments/assets/a8e9a40c-daa7-4cce-998f-94da96b7a333" />

<img width="912" height="460" alt="Character Limit when editing a channel is 24 characters" src="https://github.com/user-attachments/assets/fd09c36b-38ff-4082-ba25-f7e35b55168a" />

I suggest increasing the character limit altogether, since 16 characters (or even 24 characters) can be a bit too short

PR #182, opened 2026-02-14
Different channel name character limits when creating/editing #180

GoldcrafterXD:
Changed Character Limits to highest displayable in UI
Adresses #180 

- Categories now have a max name length of 32
- Channels have a max name length of 27 (also now equal on add and create)
- Roles have a max name length of 26
- Emojis now consistently have a max length of 32


diogomartino (approved):
LGTM

diogomartino:
My OCD kicked in a little bit when I saw you use odd numbers but it's fine :laughing:

Issue #146, opened 2026-02-13
Bug: Device options have no effect. Only top-most option used.


ghost:
At least when selecting source, settings are not actually affecting anything. Only the top-most device is selected. The settings are being saved however as disconnecting from the server and rejoining leaves the device sources the same. But changing from Microphone A to Microphone B results in Microphone A still being the source of audio.

Note, for those of us that use virtual cameras such as through OBS results, this also results in a "Could not start video source" error as those programs have exclusive control of the device and the actual webcam is at the top of the device list.

ghost:
Correction, it appears that it is not the "top-option," it just was in my test case.

From what little information I can find and understand, this appears to be due to permissions being asked to allow a specific device and not all devices of a type. So instead of "hey, can we access all your microphones?" the request is "can we access this single microphone?"

So, on some browsers this is a generic "allow/deny" pop-up and so only gives permission to the device in use when permission is given. For other browsers this is a drop-down list and a single device must be chosen. In which case, the device chosen in the prompt is the one used in channels regards of settings made in the device tab.

Perhaps there is a way to initiate the device request after every device switch within settings? Like upon hitting save if device was switched? http://discord.com/app seems to do it just fine with a single permissions request though it had trouble seeing a list of devices until I refreshed.

ghost:
Confirmed. This is an issue on chromium browsers as an **ideal** device is an overwrite by the permissions dialog. On browsers like Firefox, this is not an issue.

https://issues.chromium.org/issues/438606278

potential fix commit 4636195d1ff9fd34ba16e495cfae61416d941aef

ghost:
fix implemented #193

PR #193, opened 2026-02-14
[Bug #146] Chromium Browsers Ignores Device Selection

ghost:
Changing ideal to exact to function with chromium browsers.

**Problem:** for Chromium browsers "ideal" is now set by the user's Allow permission prompt. This results in device selection within the app being completely ignored. See: https://issues.chromium.org/issues/370332086

**Solution:** Use exact within the voice card instead.

I'm sure someone more skilled can come up with a more elegant solution such as browser detection or something but at least this can get eyes on cause of issue #146 

Edit to add > Looking around, it appears Exact actually the more appropriate argument to use anyways in this instance. Especially with the fix of #116 ?


diogomartino (approved):
LGTM

PR #204, opened 2026-02-15
feat: Add GIF search (Giphy)

ClamShoes:
Adds inline GIF search in the message composer via the emoji picker. Configurable in Server Settings; leave the API key empty to disable. Linked images (e.g. pasted or inserted image URLs) now show only the image; the URL is hidden.

**How to get a Giphy API key**
1. Go to https://developers.giphy.com/dashboard/
2. Sign up or log in and create an app
3. Copy the API key and paste it in Server Settings → General → Giphy API Key

gifsearch

### DIFF
--- a/apps/client/src/features/server/admin/hooks.ts
+++ b/apps/client/src/features/server/admin/hooks.ts
@@ -76,24 +76,37 @@ export const useAdminGeneral = () => {
     const trpc = getTRPCClient();
 
     try {
-      await Promise.all([
-        trpc.others.updateSettings.mutate({
-          name: settings.name,
-          description: settings.description,
-          password: settings.password || undefined,
-          allowNewUsers: settings.allowNewUsers,
-          enablePlugins: settings.enablePlugins
-        }),
-        trpc.others.updateGiphyConfig.mutate({
+      await trpc.others.updateSettings.mutate({
+        name: settings.name,
+        description: settings.description,
+        password: settings.password || undefined,
+        allowNewUsers: settings.allowNewUsers,
+        enablePlugins: settings.enablePlugins
+      });
+      try {
+        await trpc.others.updateGiphyConfig.mutate({
           apiKey: settings.giphyApiKey
-        })
-      ]);
-      toast.success('Settings updated');
+        });
+        toast.success('Settings updated');
+      } catch (giphyError) {
+        toast.warning(
+          'Server settings were saved, but the Giphy API key could not be updated.'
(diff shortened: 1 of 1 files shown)


toniccinot:
Squeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee!

diogomartino:
Idk about this one. While I'm against adding bloat, I understand that this is a thing that every chat app has and people are expecting it.

DrPepperG:
Why add a feature that is restricted to one provider? I think it would be a better implementation if this allowed for dynamic configuration based on what the server owner would like to use, especially since most providers have similar setups.

AbyssalSoda:
> Why add a feature that is restricted to one provider? I think it would be a better implementation if this allowed for dynamic configuration based on what the server owner would like to use, especially since most providers have similar setups.

I partially agree, restricting content to a single provider or mainstream service gives a very sanitized and limited view of whats out there. What it does provide though is convenience; with quick access to gifs most people don't readily have saved on hand.

I think a system that would allow you to see gifs set by the server owner AND your own library of gifs would be a better idea. A system could be developed to allow users to right click on gifs they come across and add them to their own library. That way gifs could be organically spread.

diogomartino:
I don't want to pollute the app with these kind of features that are really specific, because now it's giphy, tomorrow is gonna be something else. This would be an excelent use case for a plugin so server owners can choose whether if they want extra functionalities or not. Ofc the plugin SDK atm is not ready to build things like this but I'm already thinking about ways of making it possible.

In the future we might also have a marketplace for plugins so users can find them easily.

MrMightyNighty:
> I don't want to pollute the app with these kind of features that are really specific, because now it's giphy, tomorrow is gonna be something else.

I wouldn't say that GIFs should be considered as bloat. For some communities they are probably more used than emojis. You said it yourself that people are expecting it and nearly every chat app has a GIF implementation. Of course on Windows I could use `WIN + . `and paste the GIF from there but this logic would also apply to emojis.

I also agree with @DrPepperG It's probably better to implement a dynamic configuration and let the server admin decide which one they want to use.

diogomartino:
Yea, bloat is a strong word for this. But still, since there's no standard way of implementing gifs, it should be a plugin and not something sharkord ships by default.

diogomartino:
With the recent changes to plugins, this specific use case should be achievable with a plugin. Sorry about any wasted work :/

PR #208, opened 2026-02-15
Support cross-filesystem file operations

NukieGIT:
Fixes EXDEV error when uploading a file and thus allows symlinking the public folder to other locations such as an NFS share.

### DIFF
--- a/apps/server/src/utils/file-manager.ts
+++ b/apps/server/src/utils/file-manager.ts
@@ -36,6 +36,20 @@ const md5File = async (path: string): Promise<string> => {
   return hash.digest('hex');
 };
 
+const moveFile = async (src: string, dest: string) => {
+  try {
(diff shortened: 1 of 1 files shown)


diogomartino (changes requested):
That will work fine for small files, for big files it's gonna make the experience worse for everyone. I would do this:

```
import { rename, copyFile, unlink } from 'fs/promises';

const moveFile = async (src: string, dest: string) => {
  try {
    await rename(src, dest);
  } catch (err: any) {
    if (err.code === 'EXDEV') {
      await copyFile(src, dest);
      await unlink(src);
    } else {
      throw err;
    }
  }
};
```

So for 99% of the uses cases is gonna work and be fast, for your and similar use cases it's gonna fallback to the new method.

diogomartino (approved):
LGTM ty

PR #209, opened 2026-02-15
Feat( #143 #229 ) Allow User Grouping by Roles which colors Usernames

### LINKED ISSUE #143: Add Categories to User Sidebar
Adding Categories to the User Sidebar like Online and Offline and allowing Roles to Group Online Users could be a nice addition. Will try myself to add.

GoldcrafterXD:
fixes #143 #229
- Added is_grouping and order_nr fields to roles table
- Adapted Roles Screen in Server Administration to allow you to move Roles Ordering and set if Users should be grouped by the Role
- Displays Role Colors in Users Tab and Text Channels for Username

{C30C1902-8F7A-40BD-A432-89C767436762}

{7D52FD75-7D56-4A6A-8462-6056D27E6CB8}

### DIFF
--- a/apps/server/src/__tests__/seed.ts
+++ b/apps/server/src/__tests__/seed.ts
@@ -140,7 +140,7 @@ const seedDatabase = async (db: BunSQLiteDatabase) => {
     isPersistent: false,
     isDefault: false,
     createdAt: firstStart,
-    isGrouped: true,
+    isGrouping: true,
     orderNr: 2
   };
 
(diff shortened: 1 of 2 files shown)


diogomartino:
I don't like the colors, I think the default UI should be pretty neutral across the board and those colors breaks the neutrality. Also I understand the grouping, but it's not a priority at the moment and the extra complexity is not worth it, for now. Gonna close this, sorry for any wasted work you got here. In the future before making these new features, please open a issue first so we can discuss it before you jump into actual implementation. Please read: https://github.com/Sharkord/sharkord/blob/development/CONTRIBUTING.md

PR #212, opened 2026-02-15
Check for invalid Channel/Category names

### LINKED ISSUE #198: Unable to edit category after changing name to a single space.
I was trying to see how categories would look with no name, so I entered a single space character as the name for a category. 

Image

Now, I am unable to edit the category at all, as I am unable to right-click on it's name to bring up the context menu.

Herbycheck:
Since the span containing the channel's name was the trigger for the context menu, it was impossible to target it if the channel name was only one space.

I also added a check to see if a channel/category name is "empty" (only spaces). If it is empty it will display a generic "Channel" or "Category" instead of the name.
image
Let me know if this is wanted.

An alternative to this is disallowing these "empty" names server-side (when editing/creating), but this solution still allows users to have empty channel and category names should they want that for some reason.

Fixes #198


ghost:
Nice. I definitely think **both** this and a block should be implemented. In case an error or exploit ever results in a null/empty field, it's best to be able to fix it via admin controls.

Giggitybyte:
Thank you so much for this fix :slightly_smiling_face: 

If empty category names are disallowed, then I feel channels should be allowed to exist ungrouped/categorized. There are a handful of unrelated channels that I don't want to label with a category name as it'd feel forced.

Herbycheck:
@Giggitybyte Good idea, I took the liberty of creating a new issue for it, I'm not sure if I want to implement that in this PR

packages/shared/src/helpers/is-valid-input.ts
```diff
@@ -1,23 +1,23 @@
...
-    .replace(/\u00A0/g, " ")
-    .trim();
-
-  const hasText = cleaned.length > 0;
-
-  return !hasText && !hasMedia;
-};
-
-export { isEmptyMessage };
+const isValidInput = (content: string): boolean => {
```
diogomartino: This rename doesn't make sense. The function is literally checking if it's an empty message.

Your change would mean that:

isValidInput('') -> true

which doesn't make sense

apps/server/src/routers/categories/update-category.ts
```diff
@@ -12,7 +12,7 @@ const updateCategoryRoute = protectedProcedure
   .input(
     z.object({
       categoryId: z.number().min(1),
-      name: z.string().min(1).max(32)
+      name: z.string().min(1).max(32).refine((value) => !isValidInput(value))
```
diogomartino: Same as above

apps/server/src/routers/channels/add-channel.ts
```diff
@@ -13,7 +13,7 @@ const addChannelRoute = protectedProcedure
   .input(
     z.object({
       type: z.enum(ChannelType),
-      name: z.string().min(1).max(27),
+      name: z.string().min(1).max(27).refine((value) => !isValidInput(value)),
```
diogomartino: Same as above

apps/client/src/components/context-menus/channel/index.tsx
```diff
@@ -59,7 +59,7 @@ const ChannelContextMenu = memo(
       <ContextMenu>
         <ContextMenuTrigger>{children}</ContextMenuTrigger>
         <ContextMenuContent>
-          <ContextMenuLabel>{channel?.name}</ContextMenuLabel>
+          <ContextMenuLabel>{isValidInput(channel.name) ? "Channel" : channel.name}</ContextMenuLabel>
```
diogomartino: This should never be needed if the server side validation is correct, which should be where the fix is made.

Herbycheck:
closed in favor of #266

Issue #213, opened 2026-02-15
Change buttons/links to use cursor:pointer


selenecodes:
I know this is currently a hotly debated topic in the TailwindCSS / shadcn/ui space but I was wondering if it was possible to make buttons in the app use the default "cursor:pointer" behaviour that we've come to expect from the web over the years. _Or at least give the option to users*_. This is also the default behaviour currently in the discord app.

**Relevant links:**
- https://github.com/tailwindlabs/tailwindcss/issues/8961
- [Tailwind V4 upgrade-guide - Button behaviour](https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor)

**How to implement the old behaviour**
- [Tailwind V4 upgrade-guide - Button behaviour](https://tailwindcss.com/docs/upgrade-guide#buttons-use-the-default-cursor)
```css
@layer base {
  button:not(:disabled),
  [role="button"]:not(:disabled) {
    cursor: pointer;
  }
}
```

PR #214, opened 2026-02-15
fix(tailwindcss): Change button behaviour to use cursor:pointer

selenecodes:
In TailwindCSS V4 buttons were changed from the original `cursor: pointer` behaviour to `cursor: default`. Discord, which this app is based off uses the `cursor: pointer` behaviour so most users would expect a similar UX.

Linked issue: https://github.com/Sharkord/sharkord/issues/213


diogomartino:
Idk about this one, I actually like buttons not having the pointer. At most we need to make it consistent.

stereochuck:
The best practice is to be consistent with the current standard in the field.

In UX, all elements that has a click event binded (mouseOver, drag, click etc.) must have the cursor: pointer; because it defines the intent visually and instantly. The user don't have to guess what is clickable or not, which is a big plus in term of user experience even if there's an hover state on the button.

diogomartino:
Ok ok ok you have a point. I still dislike it tho :cry:

PR #223, opened 2026-02-15
Delete the message when deleting all media (if the message has no text) #192

### LINKED ISSUE #192: Delete the message when deleting all media (if the message has no text)
When someone uploads an image and/or a video and only has that attachments without any text to the message, and all media gets deleted (either by the user or those who have that permission (for an example the owner)), delete the whole message. Don't leave an empty message in the chat.

GoldcrafterXD:
fixes #192

- Added safety check to make sure message still exists in db
- Added Permission check on server side if user is allowed to manage messages or if its his message
- Added check if message contains text or other files, if not delete message aswell

NOTE: 
PR Check for invalid Channel/Category names #212
changes the function name for isEmptyMessage, if #212 will be merged this change needs to accomodate for the different Function name, please leave a comment if thats the case so i can change it


diogomartino (approved):
LGTM

PR #224, opened 2026-02-15
Add highlighting to the voice channel you're currently in #171

### LINKED ISSUE #171: Feature - Add highlighting to the voice channel you're currently in
Would be nice for the voice channel you are currently in to highlight, show an accented outline or something similar. From a UX perspective this would be really nice to have, especially when servers have large numbers of channels/users

GoldcrafterXD:
fixes #171

- Added green border around Voice Channel you are Currently in

I added a green Border around the Voice Channel you are currently in, I personally like how it looks with the Border around the Channel and that it is slightly offset to the others. However if a different look is wanted I will change it

{B62234BF-6C57-497E-87E0-8DF06A80C419}

### DIFF
--- a/apps/client/src/components/left-sidebar/channels.tsx
+++ b/apps/client/src/components/left-sidebar/channels.tsx
@@ -228,7 +228,7 @@ const Channel = memo(({ channelId, isSelected }: TChannelProps) => {
               dragHandleProps={{ ...attributes, ...listeners }}
             />
           )}
-          {channel.type === 'VOICE' && (
+          {channel.type === 'VOICE' && channel.id !== currentVoiceChannelId && (
             <Voice
               channel={channel}
               isSelected={isSelected}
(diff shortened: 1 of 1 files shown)


6004cc0-cmyk:
That looks very nice, I really like it. But if you can, add a setting to the user settings where the user can change the colour of the border.

AbyssalSoda:
VoiceChannel_Highlights
I've been working on a similar complimentary feature for voice channels as well, though not quite yet ready to pull.

GoldcrafterXD:
damn thats nice, i like yours way better xD one thing i would maybe recommend is only showing the wavy icon if you're in the channel. I think if its always shown in Voice Channels with people the UI could look a bit confusing

6004cc0-cmyk:
I agree with @GoldcrafterXD, but make the wavy icon the same width as the icons for each channel.

AbyssalSoda:
> I agree with @GoldcrafterXD, but make the wavy icon the same width as the icons for each channel.

I definitely feel like my design tastes go a bit against the grain here as I do prefer the larger icons and rounded edges of more modern applications xD.  That said, I will agree its probably more aligned with diogomartino's vision for the application.  This is what it looks like with the icon restraints. 

VoiceChannels2

6004cc0-cmyk:
Looking really good. But I don't like the eye when you share the screen. Use a desktop icon instead. And if you can, use a matching icon for what device you're using; [a smartphone icon](https://tabler.io/icons/icon/device-mobile) if you're sharing your screen from the phone, [a desktop icon](https://tabler.io/icons/icon/device-desktop) if you're on your desktop, or [a laptop icon](https://tabler.io/icons/icon/device-laptop) if you're using a laptop. Just some examples. Or better yet: just use a universal icon like [a screen share icon](https://tabler.io/icons/icon/screen-share).

Will these icons only show for you and not for the other users in the same channel?

Issue #218, opened 2026-02-15
Allow channels to exist outside of categories


Herbycheck:
[...] I feel channels should be allowed to exist ungrouped/categorized. There are a handful of unrelated channels that I don't want to label with a category name as it'd feel forced.

_Originally posted by @Giggitybyte in https://github.com/Sharkord/sharkord/issues/212#issuecomment-3904918512_

Herbycheck:
This could probably be implemented alongside #183

PR #228, opened 2026-02-16
Feature #154: Allow nested channel categories

### LINKED ISSUE #154: Feature Request: Allow nested channel categories
It would be nice to be able to put categories inside of categories for more fine grained categorization.
Something like
- Category 1
  - Subcategory 1
     - Channel
  - Subcategory 2
     - Another Channel

Maybe store a category as composites, I don't know how feasible this is with the current database

GoldcrafterXD:
fixes #154 #218

- Added category_id field to categories Table
- Added Channels to be directly created on Server without Category
- Added possibility to add Categories to Categories

Here an example
{488F0A0E-7458-40E0-A49F-BAE4EDF97CCD}


diogomartino:
This is not a priority and the designs would need to be improved in order for this to be implemented correctly. In the future please create a ticket first so we can discuss if this makes sense. Sorry!

PR #237, opened 2026-02-16
feat: implement OIDC authentication

### LINKED ISSUE #41: OIDC/LDAP/SSO support
Thanks for working on this project! I've been searching for a Discord alternative that had functional screensharing, and Sharkord seems to fit perfectly for my friend group. 

However, one thing that is missing would be support for OIDC/LDAP as the user authentication backend. I currently use Authelia for SSO/OIDC and LLDAP as the backing user directory, and since Sharkord doesn't have any support for either OIDC or LDAP, this means that my friends would need to remember a second set of credentials.

Any form of support for SSO would be much appreciated, whether that be supporting OIDC, LDAP, or trusted header auth.

zigzatuzoo:
- Added login and callback routes
- Added automatic user account creation
- Added role and group syncing
- Fixed issuer URL and SSL certificate issues

Closes #41


apps/server/src/http/oidc.ts
```diff
@@ -0,0 +1,221 @@
...
+import { eq } from 'drizzle-orm';
+import { sha256 } from '@sharkord/shared';
+import { getDefaultRole, getRoles } from '../db/queries/roles';
+import { getServerToken } from '../db/queries/server';
+import { publishUser } from '../db/publishers';
+import { getUserByIdentity } from '../db/queries/users';
+import { randomUUID } from 'crypto';
+
+// Allow self-signed certs (Delete in production if using real SSL)
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
```
Giggitybyte: Any specific reason you set this? This environment variable globally disables TLS certificate validation for the entire application. This could lead to man-in-the-middle attacks, not just for OIDC, but all HTTPS connections. It should be removed.

apps/server/src/http/oidc.ts
```diff
@@ -0,0 +1,221 @@
...
+    const as = await getOidcConfig();
+    const currentUrl = getCurrentUrl(req);
+
+    // 1. Standard Token Exchange
+    const response = await client.authorizationCodeGrant(
+        as,
+        currentUrl,
+        {
+            idTokenExpected: true,
+            expectedState: client.skipStateCheck
```
Giggitybyte: It'd be more secure to generate and store a state value in a cookie and validate it on callback to prevent token replay attacks.

apps/server/src/http/oidc.ts
```diff
@@ -0,0 +1,221 @@
...
+    // Issue App Token & Redirect
+    const token = jwt.sign({ userId: user.id }, await getServerToken(), {
+      expiresIn: '86400s' // 1 day
+    });
+
+    const clientUrl = req.headers.referer || req.headers.origin || `http://${req.headers.host}`;
+    const redirectUrl = new URL(clientUrl);
+    redirectUrl.searchParams.set('token', token);
+
+    res.writeHead(302, { Location: redirectUrl.toString() });
```
Giggitybyte: This seems like a huge security vulnerability. 

The JWT token is placed directly in the URL query string. This exposes the token in server access logs (of any proxies or CDNs) and the Referer header of subsequent requests. A malicious actor can set these headers to redirect the callback (along with the valid JWT) to their own server resulting in token theft. 

While the client does remove the token from the URL via `replaceState`, any code that runs before this (including browser extensions, analytics scripts, etc.) can capture the token from `window.location.search`. The token is also stored in `sessionStorage`, which is accessible to any JavaScript running on the same origin (XSS risk).

It should instead send an HttpOnly cookie to the client and only redirect to `config.oidc.redirectUrl`.


Something like:
```typescript
const redirectUrl = new URL(config.oidc.redirectUrl);

res.setHeader('Set-Cookie', [
  `sharkord_token=${token}; Path=/; HttpOnly; SameSite=Lax; Secure; Max-Age=86400`
]);
```

apps/server/src/http/oidc.ts
```diff
@@ -0,0 +1,221 @@
...
+        currentUrl,
+        {
+            idTokenExpected: true,
+            expectedState: client.skipStateCheck
+        },
+        {
+            redirect_uri: config.oidc.redirectUrl
+        },
+        {
+          [client.allowInsecureRequests as unknown as string]: true
```
Giggitybyte: This permits OIDC token exchanges over plain HTTP. Combined with the disabled TLS verification, this makes the entire OIDC flow trivially interceptable.

apps/server/src/http/oidc.ts
```diff
@@ -0,0 +1,221 @@
...
+    // --- USER SYNC & LOGIC ---
+
+    // Check Required Group
+    if (config.oidc.requiredGroup) {
+        const userGroups = (claims.groups as string[]) ?? (claims.roles as string[]) ?? [];
+        const hasRequiredGroup = userGroups.some(g => g.toLowerCase() === config.oidc.requiredGroup!.toLowerCase());
+        
+        if (!hasRequiredGroup) {
+            res.writeHead(403);
+            res.end(`Access Denied: Missing required group '${config.oidc.requiredGroup}'`);
```
Giggitybyte: The error response reveals the exact required group name to unauthenticated users. This leaks information that could help an attacker craft targeted privilege escalation attacks against the OIDC provider.

zigzatuzoo:
Wait ... no your completely right ... I committed the wrong version ... This was my old oidc.ts where I was trying to diagnose some IdP configuration issues. I'll send the correctly implemented (and cleaner one) here soon.

zigzatuzoo:
@Giggitybyte just pushed the actual commit.

apps/server/src/http/oidc.ts
```diff
@@ -0,0 +1,244 @@
...
+      const dbRole = allDbRoles.find(
+        (r) => r.name.toLowerCase() === localRole.toLowerCase()
+      );
+      if (dbRole) targetRoleIds.push(dbRole.id);
+    }
+  }
+
+  if (targetRoleIds.length > 0) {
+    await db.delete(userRoles).where(eq(userRoles.userId, userId));
+    await db.insert(userRoles).values(
```
Giggitybyte: It seems roles are only added if at least one OIDC group matches a mapped role. If zero groups match, the function does nothing and the user keeps whatever roles they already have.

Consider this scenario:

- A user logs in via OIDC with group `Administrator`, which maps to the local `Admin` role
- They get the `Admin` role in the Sharkord database
- Later, the OIDC admin removes the user from the `Administrator` group
- The user logs in again; their groups is now empty or doesn't include `Administrator`
- `targetRoleIds` ends up empty and this block is skipped entirely
- The user keeps their `Admin` role in Sharkord

This means role revocation at the OIDC provider is never actually propagated to Sharkord. The fix is straightforward; always sync roles when a mapping is configured, even if the result is zero roles.

However, there is another problem with this: OIDC login wipes locally assigned roles.

Every role the user has is deleted, then only the OIDC mapped roles are inserted. There's no distinction between "this role was assigned by OIDC" and "this role was assigned by a Sharkord admin manually."

The root issue is that the system doesn't track _where_ a role assignment came from. It'd be a good idea to add a `source` column to `userRoles` (e.g. 'oidc' vs 'local') and only delete/replace `source = 'oidc'` roles during sync, or maintain a separate `oidcUserRoles` table that gets synced, while leaving the main `userRoles` table alone for manual assignments.

apps/server/src/http/info.ts
```diff
@@ -15,7 +16,8 @@ const infoRouteHandler = async (
     name: settings.name,
     description: settings.description,
     logo: settings.logo,
-    allowNewUsers: settings.allowNewUsers
+    allowNewUsers: settings.allowNewUsers,
+    oidcEnabled: !!config.oidc.issuer && !!config.oidc.clientId
```
Giggitybyte: This should also take into account the value of `config.oidc.oidcEnabled`. The OIDC login button could appear on the frontend even when the admin has explicitly set `oidcEnabled: false`.

Giggitybyte:
There's one more thing I'd personally want if you're willing to implement it here: a mechanism to have clients automatically redirected to the OIDC flow for authorization instead of requiring users to click "Login with OIDC". This would allow for a seamless single sign on experience.

Perhaps a simple configuration value that does this, like `disableLocalLogin`. If `disableLocalLogin` and `oidcEnabled` are both `true`, the login page should be bypassed and the client redirected to the OIDC flow.

diogomartino:
I'll need to think about if this makes sense to add at this stage.

Giggitybyte:
Given the scope and target audience of Sharkord (friend groups and small communities that selfhost their own infrastructure) I believe OIDC authentication is particularly valuable. In the selfhosting community, it's quite common for people who run services for friends and family to also maintain a central SSO/OIDC provider (like Authentik, Authelia, or Keycloak). This allows them to manage access across all their services in one place rather than juggling separate credentials for each app.

More importantly, OIDC would help solve a practical onboarding problem for small, related communities. Consider a scenario where several friend groups each run their own Sharkord instance but occasionally want to connect across instances. With OIDC, they could agree on a common identity provider, allowing users to seamlessly join any of those instances with their existing credentials; no password resets, no remembering different logins. This would significantly lower the friction for users moving between related communities.

diogomartino:
I see, it's probably a nice to have. Not gonna merge right now because I need to study the implementation and the spec, know nothing about it.

apps/server/src/http/oidc.ts
zigzatuzoo: > It seems roles are only added if at least one OIDC group matches a mapped role. If zero groups match, the function does nothing and the user keeps whatever roles they already have.
> 
> Consider this scenario:
> 
> * A user logs in via OIDC with group `Administrator`, which maps to the local `Admin` role
> * They get the `Admin` role in the Sharkord database
> * Later, the OIDC admin removes the user from the `Administrator` group
> * The user logs in again; their groups is now empty or doesn't include `Administrator`
> * `targetRoleIds` ends up empty and this block is skipped entirely
> * The user keeps their `Admin` role in Sharkord
> 
> This means role revocation at the OIDC provider is never actually propagated to Sharkord. The fix is straightforward; always sync roles when a mapping is configured, even if the result is zero roles.
> 

That is something that I had been thinking about right before I committed it, but my thought was that they wouldn't loose the role until the next login flow ... So I was in-between revoking at login or leaving it for the user to remove...
I ended up deciding to leave it as is until I could figure out a better method of (hopefully live) updating the roles.

> However, there is another problem with this: OIDC login wipes locally assigned roles.
> 
> Every role the user has is deleted, then only the OIDC mapped roles are inserted. There's no distinction between "this role was assigned by OIDC" and "this role was assigned by a Sharkord admin manually."
> 
> The root issue is that the system doesn't track _where_ a role assignment came from. It'd be a good idea to add a `source` column to `userRoles` (e.g. 'oidc' vs 'local') and only delete/replace `source = 'oidc'` roles during sync, or maintain a separate `oidcUserRoles` table that gets synced, while leaving the main `userRoles` table alone for manual assignments.

I really do like this more detailed idea. I will start working on this flow later when I get the time.

At the moment for live role updates the only feasible idea that comes to mind is to write a bot for specifically authentik to live update the roles via using a service account but again that would only be for authentik and not for any other OIDC IdP.

Guardian259:
This is single-handedly the one feature I am waiting for. It will make the friction of migrating individuals off of public services over to an instance way easier.

zigzatuzoo:
Have most of the auth flow fixed. Just need to find some time either tomorrow or the day after to finish the rest.

diogomartino:
@zigzatuzoo so this is not fully implemented yet? In that case please change this to a draft PR so it's easier for me to keep track of things.

zigzatuzoo:
@Giggitybyte how does it look?

apps/server/src/http/oidc.ts
jlroskens: > The root issue is that the system doesn't track where a role assignment came from. It'd be a good idea to add a source column to userRoles (e.g. 'oidc' vs 'local') and only delete/replace source = 'oidc' roles during sync, or maintain a separate oidcUserRoles table that gets synced, while leaving the main userRoles table alone for manual assignments.

This seems to add unnecessary complexity to me.  I'm trying to think of a scenario where this behavior would be desired, but I can't think of one. At best it is a solution in search of a problem, at worse it's a security flaw.

If you have enabled an Identity Provider then the assumption is you will be managing claims fully through that integration. I would not want to have a different set of permissions applied outside of what has been configured at the IDP.  If they truly are a local user, then of course that makes sense, but I wouldn't appreciate having to worry about whether or not a new or existing user has roles defined outside of my identity platform.

I will say, in my experience, this is the pattern I've seen enterprise platforms use because no one wants to go digging through every app (or writing automations to do so), exploring whether or not a user has the correct permissions applied.

zigzatuzoo: > I will say, in my experience, this is the pattern I've seen enterprise platforms use because no one wants to go digging through every app (or writing automations to do so), exploring whether or not a user has the correct permissions applied.

Was thinking about this during the implementation ... I do agree. I will add a server config like `ENFORCE_OIDC_ROLES` that will allow for oidc to remove roles that are in the oidc role map json that are not given by oidc.

Giggitybyte:
@zigzatuzoo I'm not noticing any glaring issues in the code. Would you mind resolving the merge conflicts so I can pull and build this with the latest version of Sharkord to test it out?

zigzatuzoo:
Okay this should be ready to merge.

diogomartino:
I haven't had the time to look at this yet. In the meantime I'm trying code rabbit help me review PRs so I'm gonna give it a try here.

@coderabbitai review this PR. Make sure you also always check if the committed code follows the same code design as the rest of the code base to make things consistent. Save this command and use it all the time you review PRs in this repo.

NitroBrude:
> This is single-handedly the one feature I am waiting for. It will make the friction of migrating individuals off of public services over to an instance way easier.

Same, this would be huge and is the final puzzle piece needed for me to make the switch for my community. Especially if we can optionally disable/skip the local login step like @Giggitybyte brought up.

apps/server/src/http/index.ts
```diff
@@ -38,7 +39,9 @@ const routeHandlers: Partial<
   GET: {
     exact: {
       '/healthz': (req, res) => healthRouteHandler(req, res),
-      '/info': (req, res) => infoRouteHandler(req, res)
+      '/info': (req, res) => infoRouteHandler(req, res),
+      '/auth/login': (req, res) => oidcLogin(req, res),
+      'auth/callback': (req, res) => oidcCallback(req, res)
```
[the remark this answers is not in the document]
Giggitybyte: @zigzatuzoo I seem to be encountering this issue here; `/auth/callback/` returns HTTP 404, leading to an incomplete flow.

apps/server/src/http/oidc.ts
```diff
@@ -0,0 +1,284 @@
...
+    }
+    
+    const tokenResponse = await client.authorizationCodeGrant(as, url, {
+      pkceCodeVerifier: code_verifier,
+      expectedState,
+      expectedNonce,
+    });
+
+    const claims = tokenResponse.claims();
+    if (!claims || !claims.sub || !claims.email) {
```
Giggitybyte: @zigzatuzoo It seems the `/auth/callback` was not registered properly, leading to an incomplete flow. I fixed that locally and rebuilt the Docker container, but the code here is throwing.

The error occurs because the code above incorrectly expects both sub AND email in the ID token, when only sub is guaranteed to be present according to the [OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#IDToken).

I am using Authelia, which tries to follow the OIDC spec closely, and it does not provide the email claim in the JWT ID token. This information must be fetched from the UserInfo endpoint. See Authelia docs [here](https://www.authelia.com/integration/openid-connect/openid-connect-1.0-claims/), and OIDC docs [here](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo).


So, this validation should only require the sub claim, and email needs to be optional. For compatibility with other OIDC providers which don't implement the spec correctly, check to see if the email claim was included with the ID token, and call the UserInfo endpoint as a fallback.

apps/server/src/http/oidc.ts
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    if (config.oidc.requiredGroup) {
+      const groups = (claims.groups as string[]) || [];
+      if (!groups.some(g => g.toLowerCase() === config.oidc.requiredGroup?.toLowerCase())) {
+        return res.writeHead(403).end('Forbidden');
+      }
+    }
+
+    const identity = (claims.email || claims.sub) as string;
```
Giggitybyte: Also just noticed this: using `email` is not a smart idea to uniquely identify a user, only `sub` should be used. If a user changes their email address in their IDP, they'd lose access to their Sharkord account despite logging in with the same IDP user account.

From the [OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#ClaimStability):
> ... other Claims such as email, phone_number, preferred_username, and name MUST NOT be used as unique identifiers for the End-User, whether obtained from the ID Token or the UserInfo Endpoint.

apps/server/src/http/oidc.ts
```diff
@@ -0,0 +1,284 @@
...
+    
+    // Set OIDC session cookie
+    res.setHeader('Set-Cookie', `__Host-oidc_session=${encodeURIComponent(sessionData)}; HttpOnly; Secure; SameSite=Lax; Max-Age=300; Path=/`);
+
+    const baseUrl = getBaseUrl(req);
+    const redirectUri = `${baseUrl}/auth/callback`;
+
+    const parameters: Record<string, string> = {
+      redirect_uri: redirectUri,
+      scope: 'openid profile email groups',
```
Giggitybyte: Important to note: `groups` is not a standard scope. While I believe most IDPs use `groups`, some use `roles` or `memberOf`, and others use even more non-standard stuff like nested objects (e.g. `resource_access.client_name.roles`).

There should be a configuration value that allows the end-user to specify the name of the scope/claim for groups, with `groups` as the default.

apps/server/src/http/oidc.ts
```diff
@@ -0,0 +1,284 @@
...
+  if (!user) {
+    const defaultRole = await getDefaultRole();
+    if (!defaultRole) throw new Error('Default role missing');
+
+    const randomPassword = createHash('sha256').update(randomBytes(32).toString('hex')).digest('hex');
+    
+    const [insertedUser] = await db.insert(users).values({
+      identity: identity,
+      password: randomPassword,
+      name: (claims.name as string) ?? identity.split('@')[0],
```
simcop2387: Along with all the other comments about the different claims needing to be configurable, this one likely should be too.  with my current keycloak setup this would end up having all users use their full name as the name, i.e. in my case `Ryan Voots` rather than the name i'd likely want here, `simcop2387` which is provided under the `preferred_username` claim by default.   Also this assumes that the identity is an email address and can be split like that, which may or may not be the case even with the existing code.  And it'll mean that two users could have the same name, i.e. `john@example.com` and `john@otherdomain.com` which is likely to cause some confusing issues.

Giggitybyte: Very good catch. Authelia works the same way; `name` would be the full given name of a user and `preferred_username` is the actual login name for a user, which would be what their Sharkord username should be. An email should never be used.

zigzatuzoo:
Okay, will have some time later today to work on all this. Thank you all for catching these issues!

univeous:
Great work! I merged and tested this PR on my branch. It works really well. Thanks for your effort!

A few minor things I noticed:

- If OIDC is the only login method and the user manually clicks "Disconnect", they get redirected back to the login page which then automatically redirects to OIDC login again. This behavior feels off. Maybe we need some kind of "manual logout" flag to prevent it.
- After a successful OIDC login, the login(callback) page still briefly appears before redirecting to the main page; this feels a bit jarring. Perhaps we could show something like "Completing OIDC login..." in the login form instead of continuing to display the OIDC login button to avoid confusing.

simcop2387:
> Great work! I merged and tested this PR on my branch. It works really well. Thanks for your effort!

>     * If OIDC is the only login method and the user manually clicks "Disconnect", they get redirected back to the login page which then automatically redirects to OIDC login again. This behavior feels off. Maybe we need some kind of "manual logout" flag to prevent it.

This should be handled by a back channel logout process when supported by the oidc provider.  This is the exact spec so it's basically useless for understanding it but it's a good reference at least, https://openid.net/specs/openid-connect-backchannel-1_0.html . But basically the idea is that Sharkord will use the auth token that it has to call an endpoint on the OIDC provider to do the logout, resulting in the user session being logged out for that oidc client on the provider.  There's also a front-channel logout where you'd redirect the user to a url on the provider to do the same thing but back-channel is preferred because it means that you're not relying on the client to do the actual logout and not just pretending to have done it and then using the token anyway.

>     * After a successful OIDC login, the login(callback) page still briefly appears before redirecting to the main page; this feels a bit jarring. Perhaps we could show something like "Completing OIDC login..." in the login form instead of continuing to display the OIDC login button to avoid confusing.

Yea usually there's a different interstitial page for the callback that does exactly that.

simcop2387:
I'll probably set this up myself tonight and play with it, should hopefully find any other potential issues with it and i'll leave more reviews/comments as necessary.  Might take a look at dealing with the logout stuff to enhance that for doing oidc session logouts with both the back-channel setup and front-channel versions for better security (it helps prevent token theft and session hijacking on the oidc provider).

Guardian259:
Eagerly Awaiting this PR!!! :eyes:

diogomartino:
I'm sorry, this has been here for a long time and I haven't looked into it yet.

Giggitybyte:
@zigzatuzoo Hello, could you resolve the merge conflicts so I can pull this PR with the latest Sharkord changes?

J-eremy:
This feature would be an absolutely amazing addition.

I just installed Sharkord last night and i just want to say you do great work. Its so simple and clean. Bravo.

zigzatuzoo:
> @zigzatuzoo Hello, could you resolve the merge conflicts so I can pull this PR with the latest Sharkord changes?

Yeah I can get to it in a bit...

diogomartino:
Guys, I'm really sorry I haven't looked at this yet. Lately I'm not having too much time to dedicate to Sharkord.

HankLi0130:
Hi @zigzatuzoo, @univeous, and @simcop2387 — I’m running a small research study on identity integration in self-hosted OSS. This PR is a useful real-world case: OIDC was implemented and tested for an internal deployment, but integration/review has continued while the branch drifted and maintainer time became scarce. If any of you worked on this for an organizational internal-messaging environment, would you be open to a 20-minute research call about the work required, security/operational consequences, and how external integration help would be prioritized or approved? This first conversation is research, not a sales pitch; a simple “interested” reply is enough.

Giggitybyte:
5 months :weary:

diogomartino:
Sorry that took so long. The main thing is: OIDC is landing as a feature but group and role syncing is not. I want to keep that exclusive to Sharkord. To be clear, this is something that I personally won't use and I'm open to discuss it later, but the first version isn't gonna have it.

Thanks for your work @zigzatuzoo and for @Giggitybyte review. I'm gonna close this PR and open a new one, but my work was heavily influenced by what I've seen here.

PR #245, opened 2026-02-16
feat: Member coloring & ordering by role

### LINKED ISSUE #143: Add Categories to User Sidebar

schemen:
This will resolve #143.

* This is currently used to add coloring for usernames by user role. For now, it prioritizes which color to use by amount of permissions of a role or as fallback by index of the role.
* The member sidebar also sorts by role priority. Members that are offline are in a separate tab without any role ordering, relying on the standard ordering already in the project.

grafik

### DIFF
--- a/apps/client/src/components/right-sidebar/index.tsx
+++ b/apps/client/src/components/right-sidebar/index.tsx
@@ -1,11 +1,24 @@
 import { UserName } from '@/components/user-name';
 import { UserAvatar } from '@/components/user-avatar';
+import { useRoles } from '@/features/server/roles/hooks';
 import { useUsers } from '@/features/server/users/hooks';
 import { cn } from '@/lib/utils';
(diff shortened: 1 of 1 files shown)


schemen:
I also noticed PR #209 - though I suppose the implementation is different, I was a bit slower :see_no_evil:

diogomartino:
Please read https://github.com/Sharkord/sharkord/pull/209#issuecomment-3922711542. I hate to close peoples PR's when they took their time to contribute but the reality is that if I don't do that the project starts losing focus. The grouping on the right is a possibility for the future but making that would also require the need of controlling the order which in itself will add unnecessary complexity at this stage.

schemen:
That's fair. Maybe I can put that stuff into a plugin :).

Issue #194, opened 2026-02-14
Add channel's topic on top of the whole chat/call


6004cc0-cmyk:
Show the topic as a glued bar on top of the whole chat (think `position: fixed;`) for all chats and calls that has a topic.

diogomartino:
I didn't even remembered that channels had topics. But I agree they should show

6004cc0-cmyk:
They have topics :) When you edit a channel, you will have 2 text fields: `Name` and `Topic`.

PR #251, opened 2026-02-16
Feat #101 Add pinned messages

### LINKED ISSUE #101: Pin message
Pinned message should show on top of the channel.

Pesuri:
## Summary

Closes #101

---

## Additional Context

Added feature to pin messages.

kuva
kuva

### DIFF
--- a/.github/actions/build-sharkord/action.yml
+++ b/.github/actions/build-sharkord/action.yml
@@ -5,7 +5,7 @@ inputs:
   bun_version:
     description: Bun version to install
     required: false
-    default: 1.3.5
+    default: 1.3.10
   bump:
     description: Version bump mode (none, patch, minor, major)
     required: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Setup Bun
         uses: oven-sh/setup-bun@v1
         with:
-          bun-version: 1.3.5
+          bun-version: 1.3.10
 
       - name: Install dependencies
         run: bun install
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -16,7 +16,7 @@ jobs:
       - name: Setup Bun
         uses: oven-sh/setup-bun@v1
         with:
-          bun-version: 1.3.5
+          bun-version: 1.3.10
 
       - name: Install dependencies
         run: bun install
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM oven/bun:1.3.5
+FROM oven/bun:1.3.10
 
 ARG TARGETARCH
 ENV RUNNING_IN_DOCKER=true
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
   [![Downloads](https://img.shields.io/github/downloads/Sharkord/sharkord/total)](https://github.com/Sharkord/sharkord/releases)
   [![Last Commit](https://img.shields.io/github/last-commit/Sharkord/sharkord)](https://github.com/Sharkord/sharkord/commits)
   
-  [![Bun](https://img.shields.io/badge/Bun-v1.3.5-green.svg)](https://bun.sh)
+  [![Bun](https://img.shields.io/badge/Bun-v1.3.10-green.svg)](https://bun.sh)
   [![Mediasoup](https://img.shields.io/badge/Mediasoup-v3.19.11-green.svg)](https://mediasoup.org)
 </div>
 
--- a/apps/client/src/components/channel-view/text/renderer/index.tsx
+++ b/apps/client/src/components/channel-view/text/renderer/index.tsx
@@ -1,6 +1,8 @@
+import { RelativeTime } from '@/components/relative-time';
 import { requestConfirmation } from '@/features/dialogs/actions';
-import { useOwnUserId } from '@/features/server/users/hooks';
+import { useOwnUserId, useUserById } from '@/features/server/users/hooks';
 import { getFileUrl } from '@/helpers/get-file-url';
+import { getRenderedUsername } from '@/helpers/get-rendered-username';
 import { getTRPCClient } from '@/lib/trpc';
 import { cn } from '@/lib/utils';
 import {
@@ -8,6 +10,7 @@ import {
   isEmojiOnlyMessage,
   type TJoinedMessage
 } from '@sharkord/shared';
+import { Tooltip } from '@sharkord/ui';
 import parse from 'html-react-parser';
 import { memo, useCallback, useMemo } from 'react';
 import { toast } from 'sonner';
@@ -23,6 +26,7 @@ type TMessageRendererProps = {
 
 const MessageRenderer = memo(({ message }: TMessageRendererProps) => {
   const ownUserId = useOwnUserId();
+  const editedByUser = useUserById(message.editedBy ?? -1);
   const isOwnMessage = useMemo(
     () => message.userId === ownUserId,
     [message.userId, ownUserId]
@@ -88,6 +92,28 @@ const MessageRenderer = memo(({ message }: TMessageRendererProps) => {
         )}
       >
         {messageHtml}
+        {message.editedAt && (
+          <Tooltip
+            content={
+              <div className="flex flex-col gap-1">
+                <RelativeTime date={new Date(message.editedAt)}>
+                  {(relativeTime) => (
+                    <span className="text-secondary text-xs">
+                      {editedByUser
+                        ? getRenderedUsername(editedByUser)
+                        : 'Unknown User'}{' '}
+                      {relativeTime}
+                    </span>
+                  )}
+                </RelativeTime>
+              </div>
+            }
+          >
+            <span className="msg-edit ml-1 text-xs text-muted-foreground">
+              (edited)
+            </span>
+          </Tooltip>
+        )}
       </div>
 
       {allMedia.map((media, index) => {
--- a/apps/client/src/components/server-screens/user-settings/devices/index.tsx
+++ b/apps/client/src/components/server-screens/user-settings/devices/index.tsx
@@ -2,6 +2,7 @@ import { useDevices } from '@/components/devices-provider/hooks/use-devices';
 import { getVoiceControlsBridge } from '@/components/voice-provider/controls-bridge';
 import { closeServerScreens } from '@/features/server-screens/actions';
 import { useCurrentVoiceChannelId } from '@/features/server/channels/hooks';
+import { usePublicServerSettings } from '@/features/server/hooks';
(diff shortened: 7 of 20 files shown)


diogomartino:
I think this is too complex. It should only show one pinned message and it should always show it on the top to get the user attention, that's the main point of the pinned message. The easiest way would be to add a `pinnedMessageId` to the channel.

Look at how whatsapp does it:

image

stb109:
I think allowing multiple pins makes sense. Sometimes we want to pin multiple links or information that can easily be revisited in a chat. We couldn't do that with only one pin.
Maybe make a announcement type pin that always shows? This could a single pin selected from the pins in the current chat.

Pesuri:
I see. I was thinking pinning messages like in Discord.
Whatsapp way feels more like an announcement and it might be better to skip pinning a message and add it straight to channels table as announcement column. It could be added and modified in channel settings.

diogomartino:
I actually didn't know how discord handled pinned messages but for me the pinning of a message was exactly that, an announcement. The way discord works is more of like bookmarks or favorites because you don't see them until you open the menu. Idk, I'll wait for more feedback on this.

DrPepperG:
> I actually didn't know how discord handled pinned messages but for me the pinning of a message was exactly that, an announcement. The way discord works is more of like bookmarks or favorites because you don't see them until you open the menu. Idk, I'll wait for more feedback on this.

Discord’s method of pinning is quite a great feature to have in text channels.

Personally I think a channel description would suffice for your sticky annnouncement idea.

Pesuri:
I made a small addition and now text from topic column from channels-table is shown on the top of the channel if it isn't empty.
kuva

Pesuri:
Or should we create a new column to channels table instead of using topic column?

ThingsCouldGetDicey:
just to put in my two cents. i think always displaying a pinned message at the top clutters the view of the chat. if we could have a way to pin and recall those messages outside of the chat. it would look neater. altho i do like something in the chat that always displays, that could be useful for like rules, or a link thats evergreen. i guess you would want to differentiate between bookmarks, and pinned messages. and do a seperate feature for both. or do what discord does which is have a pin icon that can recall all the pinned messages like bookmarks. 

the idea of having something always displaying at the top is new. but at the same time it would be an eye sore.

Aasikki:
Discords way of doing pins like bookmarks is imo a very useful and a must have feature! I constantly use it to mark important messages so people can find them later.

The announcement style pin also sounds like a useful idea, but that's imo a different feature entirely.

Calling them bookmarks instead of pins might makes sense, but then again people might hear "bookmark" and think it's like browser bookmark, so only for themselves to see, when in reality it would be public to everyone, so maybe it's best to just copy discord and call them pins to avoid confusion 🤔

Pesuri:
I made few improvements to the pr.

So currently this pull request contains both Discord like pinned messages and Channels topic as permanent message.
How should we proceed?  Should we use both? Should we make some more changes?

diogomartino:
@Aasikki I think we could have both tbh. Like a pinned message that always shows on the top (kinda agree with the eye sore aspect, but that's kinda the point: to draw attention) and favorited/bookmarked messages which should be private to each user.

@Pesuri let's wait about some other people feedback before commiting to a solution

stb109:
I like the functionality of a channel topic and multiple pinned messages.

ThingsCouldGetDicey:
> @Aasikki I think we could have both tbh. Like a pinned message that always shows on the top (kinda agree with the eye sore aspect, but that's kinda the point: to draw attention) and favorited/bookmarked messages which should be private to each user.
> 
> @Pesuri let's wait about some other people feedback before commiting to a solution

wouldnt you also want globally pinned messages? rather than it private to each user? i mean you could have both again. 

so you could have a bookmark feature of important messages per user (which doesnt currently exist on discord and would be a cool feature)

pinned messages by the server admin for people to globally see. 
a channel topic pin. which i think is a good idea. if the ui is done in a way thats less offensive.  because a common problem with discord is people dont respect the topics in selected named topic areas. it could serve as a reminder. 
or as a rules link. or something else.

diogomartino:
@stb109 I didn't even think about the channel topic. Channels already have topics but they are not shown anywhere, my concept of pinned message might just be the topic after all.

I think it should look like this:

image

On text channels it should show the header with the channel topic and on the right a pin that opens a popover with the pinned messages

stb109:
Yeah, that looks good to me.

diogomartino:
That will also solve this https://github.com/Sharkord/sharkord/issues/194

caellach:
Discord has both Channel Topic + Pinned Messages and I think they serve slightly different use cases. Channel topics tend to be for important announcements/rules whereas pinned messages tend to be for short-term group information.

Glad to see both are likely to make it in.

Pesuri:
Alright. I made some more improvements.
- Announcement was renamed as topic
- Pinned messages view box max size was increased
- Pinned messages button has been moved to Topic box as suggested
- Pinned message MOVE-button style matches with other buttons
- Pinned message MOVE-button will now center chat to pinned message, close pinned messages view box and highlight message for few seconds
- Fixed new conflicts

kuva

diogomartino:
The design is way off, try this:
- the bar shouldn't be floating
- the bar where the channel topic is should have the same height of the section that says "Sharkord Server" and "Members". It also should have a border bottom similar to those.
- the background color should be the same as the sidebars
- the pin button is too big

Pesuri:
Finally some directions. I will do those adjustments.

Pesuri:
Better?
kuva

diogomartino:
Much better. What is that "MOVE" thing?

Pesuri:
It moves view to that pinned message like JUMP button in discord.

I made a shortclip to demonstrate how it works.

Pesuri:
[demo.webm](https://github.com/user-attachments/assets/43962220-ade7-47dc-b4b4-4ecca35f3439)

diogomartino:
Nice, but I would use an icon. Also, that's not gonna work out of the box for messages that aren't loaded yet. I'm gonna review your PR probably tomorrow, thank you!

apps/server/src/routers/messages/toggle-message-pin.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    invariant(message, {
+      code: 'NOT_FOUND',
+      message: 'Message not found'
+    });
+
+    await db
+      .update(messages)
+      .set({ pinned: !message.pinned, updatedAt: Date.now() })
+      .where(and(eq(messages.id, input.messageId)));
```
diogomartino: Useless `and`

apps/client/src/components/channel-view/text/messages-group.tsx
```diff
@@ -7,55 +8,113 @@ import {
...
+    const firstMessage = group[0];
+    const user = useUserById(firstMessage.userId);
+    const date = new Date(firstMessage.createdAt);
+    const isOwnUser = useIsOwnUser(firstMessage.userId);
+    const isDeletedUser = user?.name === DELETED_USER_IDENTITY_AND_NAME;
+    const pinnedMessageContext = useContext(PinnedMessageContext);
+
+    if (!user) return null;
+
+    function scrollToMessage(id: number) {
```
diogomartino: Use ES6 styntax.

apps/client/src/components/channel-view/text/messages-group.tsx
```diff
@@ -7,55 +8,113 @@ import {
...
+      const el = messageRefs?.current[id];
+      if (el) {
```
diogomartino, line 40: Can't these styles be set directly in the component with tailwind? I don't like to mess with the dom directly unless absolutely necessary. For example you can set a flag in the store called "scrollingToMessageId" and then style the message accordingly.

apps/client/src/components/channel-view/voice/hooks/use-voice-refs.ts
diogomartino, line 1: This file doesn't need to be touched at all. Those voids are useless.

apps/client/src/components/pinned-message-provider/index.tsx
diogomartino, line 1: You should always avoid the use of contexts. In this case is not needed at all. Always use the store to store state.

apps/client/src/components/pinned-messages-box/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+type TPinnedMessagesBoxProps = {
+  className?: string;
+  isOpen?: boolean;
+  messageRefs: React.RefObject<Record<number, HTMLDivElement | null>>;
+  messages: TJoinedMessage[];
+};
+
+const PinnedMessagesBox = memo(
+  ({ isOpen = true, messageRefs, messages }: TPinnedMessagesBoxProps) => {
+    const pinnedMessages = messages.filter((msg) => msg.pinned);
```
diogomartino: Memoize with `useMemo`

apps/client/src/components/channel-view/voice/hooks/use-voice-refs.ts
Pesuri: Seems to be something that format:check wanted to be done. Well, I will remove these changes.

apps/server/src/routers/messages/toggle-message-pin.ts
Pesuri: Removed `and`

apps/client/src/components/channel-view/text/messages-group.tsx
Pesuri: Fixed to use ES6

apps/client/src/components/channel-view/text/messages-group.tsx
Pesuri: Fixed to use tailwind

apps/client/src/components/channel-view/voice/hooks/use-voice-refs.ts
Pesuri: Removed voids

apps/client/src/components/pinned-message-provider/index.tsx
Pesuri: Context has now been replaced with app slice.

apps/client/src/components/pinned-messages-box/index.tsx
Pesuri: Memoized

Pesuri:
> Nice, but I would use an icon. Also, that's not gonna work out of the box for messages that aren't loaded yet. I'm gonna review your PR probably tomorrow, thank you!

Replaced MOVE button with icon button.
It is now arrow with "Jump to message" title.
image

Pesuri:
And of course I forgot to push Move button change commit to Github yesterday. Well. Now it is here.

diogomartino:
Cool. There are still some things that need a little tweak but I'll do it myself, I want to merge this today. Thanks!

Issue #235, opened 2026-02-16
Feature: Add ability to assign audio output devices


toby-short:
I think as a start, the ability to select the audio output device is important. But as an addition, I think it would it would be **great** if we can potentially assign streams to a different audio device, or be able to right click a user or their stream and tell it to output through a different device.

A good example is say you're in a call with some friends watching a movie or playing something like jackbox games together, you want to hear the stream audio but don't want it overpowering your voice chat, directing the stream audio to a different device would solve this issue and is something I've wanted in Discord for a long time.

I've done some looking and it shouldn't be too difficult to set the audio tag output sink to a different one.

I'll probably end up looking at doing this if it's something people will want :)

Aasikki:
I think this would be a cool option to have.

AzzySama:
Yes please this would solve many issues

diogomartino:
It's done already.

Issue #80, opened 2026-02-11
Feature request: Voice activity detection


Aasikki:
Someone already requested push to talk, but imo this would be just as important, if not more so. Add a configurable thresold to the device/audio settings which the user can use to calibrate the thresold when their mic will be open.

This is usually plenty enough to cut away all background noise without affecting sound quality. The noise suppression we currently have seems to be hit or miss, it removed background noise quite well for one user, but for another user it seems to do nothing. Voice activity detection would totally solve this and give us the peace and quiet we need when no one's talking!

PR #253, opened 2026-02-16
Feat/audio and video testing

### LINKED ISSUE #252: [Feature]: audio and video preview
### Problem

We should be able to test audio and video before joining a voice channel

### Proposed Solution

Allow to test video and audio in the settings window

schemen:
## Summary

* Added microphone and speaker testing in Settings -> Devices with explicit mic permission handling.
* Implemented mic test controls (Start/Stop Test), live voice level meter, and delayed loopback playback.
* Added persistent output-device selection and routed playback/test audio to the selected speaker when supported.
* Improved device list population/refresh behavior around permissions and hardware changes.
* Added webcam testing with Start/Stop Video Preview
* Added camera startup resilience, so preview is more reliable on slower webcam initialization.
* When starting mic test during an active call, the app now mutes/deafens and then restores the previous voice state after testing.

A small side note: Added a small controls-bridge to expose minimal voice-control actions across component boundaries, so settings can trigger in-call mute/deafen behavior without deep refactors.

The bridge can be removed if this is unwished though, we'll lose the mute/unmute ability, but I think that would be okay.

Closes #252, parts of #235

---

## Additional Context

Tested it on my instance and so far it seems fine and working. 

This would be a pre-requisit for #80 

image


schemen:
Improved the audio level a bit

image


image

diogomartino:
Lot of code to review, gonna need some time to review and test, but it looks promising. Please resolve the conflict, ty!

schemen:
Upstream code is rebased and merged into the current works :)

apps/client/src/components/channel-view/voice/hooks/use-voice-refs.ts
```diff
@@ -110,7 +113,8 @@ const useVoiceRefs = (
     }
 
     audioRef.current.volume = userVolume / 100;
-  }, [audioStream, audioRef, userVolume]);
+    void applyAudioOutputDevice(audioRef.current, devices.playbackId);
```
diogomartino, line 116: No need for `void`

diogomartino (commented):
It needs some tweaks but overall I think it's good. I'll do them myself, ty!

image

The code is a bit inefficient, when you are trying the mic, while you are talking the cpu is pegged at 100%.

schemen:
> It needs some tweaks but overall I think it's good. I'll do them myself, ty!
> 
> image
> The code is a bit inefficient, when you are trying the mic, while you are talking the cpu is pegged at 100%.

Ah interesting. I have not noticed the performance on my workstation! I‘ll gladly also check how to optimise :) I think it may be related to the drawing of the audio level.

PR #254, opened 2026-02-16
Feature #236: Moderate Usernames

### LINKED ISSUE #236: Feature: Moderate Usernames
Could really use the ability to change and lock a user's username so users with the correct permissions can change problematic usernames and prevent the offenders from changing it themselves without admins unlocking.

GoldcrafterXD:
fixes #236

- Added locked_username to users table, queries and routes
- Added Lock Username / Unlock Username Button to User Management Tab with handler
- Adapted updateUserRoute to allow partial setting of User values
- Added USERNAME_LOCK ActivityLogType
- Added data_test to .gitignore

# Overview

{4E4341E4-167A-4E88-A651-3B8932D24F39}

Added Lock Username Button that becomes Unlock Username when unlocked

{65720124-B491-428A-A8E4-6F454AA1B9FB}

For Users with Permission "Manage Users" the Username is editable and will persist on Enter Press

{89F56511-1223-4E84-B2D0-9CAFB458F9C8}

Confirmation Message if you want to Lock / Unlock the Users name

{AAD46FB6-D3E6-47FA-AF0E-397148D1D4CB}

If changing Username when Locked you will get an Error Popup that your Username cannot be changed

{E352C03C-0996-48DF-A075-C5D14394E594}

But Changes to Bio, Color and Banner are still persisted

### DIFF
--- a/apps/client/src/components/mod-view-sheet/header.tsx
+++ b/apps/client/src/components/mod-view-sheet/header.tsx
@@ -7,7 +7,7 @@ import {
   requestTextInput
 } from '@/features/dialogs/actions';
 import { useUserRoles } from '@/features/server/hooks';
-import { useOwnUserId, useUserStatus, useIsOwnUser } from '@/features/server/users/hooks';
+import { useOwnUserId, useUserStatus, useIsOwnUser, useUserById } from '@/features/server/users/hooks';
 import { getTrpcError } from '@/helpers/parse-trpc-errors';
 import { getTRPCClient } from '@/lib/trpc';
 import { DELETED_USER_IDENTITY_AND_NAME, UserStatus, Permission, isEmptyMessage } from '@sharkord/shared';
@@ -18,6 +18,7 @@ import { Dialog } from '../dialogs/dialogs';
 import { RoleBadge } from '../role-badge';
 import { useModViewContext } from './context';
 import { useCan } from '@/features/server/hooks';
+import { useRoles } from '@/features/server/roles/hooks';
 
 const Header = memo(() => {
   const ownUserId = useOwnUserId();
@@ -35,10 +36,15 @@ const Header = memo(() => {
     setName(user.name);
   }, [user.name]);
 
-  const canManage = useMemo(
+  const canSetUsername = useMemo(
     () => can(Permission.MANAGE_USERS) || (isFromOwnUser && !user.lockedUsername),
     [can, isFromOwnUser, user.lockedUsername]
   );
+
+  const changedUser = useUserById(user.id);
+  const allRoles = useRoles();
+  const roles = changedUser!.roleIds.map(roleId => allRoles.find(role => role.id === roleId));
+  const canLockUsername = roles.some(role => role?.permissions.includes(Permission.MANAGE_USERS)) ? true : false;
 
   const onChangedUsername = useCallback(
     async (userId: number, newName: string) => {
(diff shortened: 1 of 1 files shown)


ghost:
So cool! Would it be too difficult to address #257 with this same solution as well?

michaelmoskie:
Would love for this to get in to main. Good work!

diogomartino:
Your implementation has lots of problems. And I don't really like the concept of locking the username. It's a really specific thing. Next we gonna need block avatar, block description and so on. It would be better to have a new permission called UPDATE_PROFILE and if you don't want a specific user to be editing their profile just remove that permission from them.

Also being able to edit a user's profile as admin is good (not just their username). If you do this, try to reuse components as much as possible.

ghost:
Is the issue that it's too specific? If so, I agree. Is the issue that it shouldn't be a moderation action? Hard disagree.

Discord has permissions for Speaking and Sending messages...**AND** a moderation option in the context menu to Mute a user server-wide.

I feel like Sharkord not only needs that function as well but also needs to do the same here since there's no validation and no central hub to report accounts to. That is, have permissions for editing profiles (both own and others)...**AND** an option to lock/unlock a profile as a moderation action that supercedes permissions.

diogomartino:
The thing to think about here is the scale. Discord actually needs to have a really good moderation system because it's build to handle huge public communities with lot's of traffic. Sharkord's main target is private server for small groups of friends. Because of this, moderation should not be a priority at the moment because at that scale you will have pretty much nothing to moderate. So my suggestion to add an extra role to an already existing system is the quickest way to achieve some more control. I'm not saying that adding more moderation won't happen, it just doesn't need to happen now.

Ruchian:
I think this feature is overreach.
You want to call someone else a name, and then force others to call said user by that same name. That's a form of crowd bullying. All that would do is cause rage and the user to either get banned for being angry and saying things, or for leaving themselves. It would turn your community into a toxic heaven for those seeking power essentially.

There are other ways to deal with names you find problematic.

I think the best option is to have a list of words you don't want to see in people's names, and then should they have said name, they will only see one or two channels and their membership would be hidden from your member list. The channels are there to explain why they only see those channels, that it has to do with their username.
This 'lock' would also apply to someone who changes their name to such username, and would be removed if they pick an acceptable name automatically.
This would prevent the need from moderating it at all, and cause a lot less other problems, such as given roles being lost or whatever.

Editted:
You basically want to avoid upsetting people at all or needing to deal with apology letters, but I'll elaborate my point here: https://github.com/Sharkord/sharkord/discussions/361

diogomartino:
I'm gonna close this. This will be revisited in the future.

I opened a discussion here: https://github.com/Sharkord/sharkord/discussions/369. Also check @Ruchian discussion since it goes into a lot of detail: https://github.com/Sharkord/sharkord/discussions/361.

ghost:
> I'm gonna close this. This will be revisited in the future.
> 
> I opened a discussion here: #369. Also check @Ruchian discussion since it goes into a lot of detail: #361.

Yikes, I think #361 proves exactly WHY I felt the need to block said individual. Odd that this is the sort of thing you want to endorse.

Ruchian:
> Yikes, I think https://github.com/Sharkord/sharkord/discussions/361 proves exactly WHY I felt the need to block said individual.

For the record, I created that discussion topic many hours after you had already blocked me, and more than 12 hours after I noticed I could no longer reply to the nonsense you posted in the proposal issues you created. My thread being the reason can therefore not be the reason for you to block me. 

If you want to be an ass without reprecussions, then sorry. Find another towel to walk over. 
Also, could you stop being vague? You literally explained nothing nor pointed out anything concrete. I don't mind talking with you or giving you some otherattention, so long the goal is to be helpful to the project and its users. You're also free to comment on my thread instead, considering your post is entirely about that. Putting your complaint here is off topic.

Issue #76, opened 2026-02-11
Request: Improve password/account security


TheraNinjaCat:
This project looks really good, however after digging into how the passwords are dealt with, I believe there could be some improvements:

- Replace SHA256 hashing with a better password hashing algorithm such as Argon2, Bcrypt/Scrypt, or PBKDF2. Explanation here: https://dev.to/lovestaco/hashing-passwords-why-md5-and-sha-are-outdated-and-why-you-should-use-scrypt-or-bcrypt-48p2
https://github.com/Sharkord/sharkord/blob/6979e48feccac79bd8778f239fea3cc9ef75d7f7/apps/server/src/http/login.ts#L31
- Add MFA options/enforcement such as TOTP and FIDO2
- Do not allow users to save their password into local storage, at most this should remember usernames.
https://github.com/Sharkord/sharkord/blob/6979e48feccac79bd8778f239fea3cc9ef75d7f7/apps/client/src/screens/connect/index.tsx#L85-L88 
  - A slightly more robust remember-me functionality may be found by using or copying Oauth2's refresh token system. A different issue, #41 mentions the use of [Passport ](https://www.passportjs.org/) which may provide this functionality, however I am unfamiliar with it. Do consider security concerns around compromise of this token itself, and potential strategies to mitigate it such as DPoP and ensuring it's one-time-use, more info here: https://oauth.net/2/refresh-tokens/
- Consider adding a forgotten password flow to allow users to regain access to the server without having to bother the owner/admin
- Force a user to reauthenticate (i.e. require a password check) before allowing them to change their username to prevent account takeover/lockouts (if a forgotten password email function is added, the email should also require a password to change)
- Increase the minimum password length, potentially allow configuration for password strength requirements.
  - Consider using a strength checker like [zxcvbn](https://www.npmjs.com/package/zxcvbn)
  - Consider using [Have I Been Pwned's api](https://haveibeenpwned.com/) to check passwords against known compromise lists (https://www.npmjs.com/package/hibp)

diogomartino:
Yea, you are right. The password handling was something that I had on my mind that I needed to improve but I never did. Now it's gonna be tough to change, specially the hashing part.

TheraNinjaCat:
[...]

PR #271, opened 2026-02-17
fix(#76): Removed local storage of sensitive credentials added argon2 bun native password hasing.

ELLEation:
## Summary

This PR removes local storage used to store sensitive credentials. Both server and user credentials. This is better left to the browser or user's password manager. The reasoning is this leaves users vulnerable to XSS attacks and is generally considered insecure for secrets.

This also adds argon2 password hashing in place of sha256

Closes #
https://github.com/Sharkord/sharkord/issues/76

---

## Additional Context

This code is tested and working on Debian 13 with the latest version of Bun and passed prerequisite checks.

I have plans to hopefully change token auth to utilize cookies instead of session storage.

### DIFF
--- a/apps/client/src/screens/connect/index.tsx
+++ b/apps/client/src/screens/connect/index.tsx
@@ -9,6 +9,7 @@ import {
 } from '@/helpers/storage';
 import { useForm } from '@/hooks/use-form';
 import { PluginSlot } from '@sharkord/shared';
+import { PluginSlotRenderer } from '@/components/plugin-slot-renderer';
 import {
   Alert,
   AlertDescription,


TheraNinjaCat:
That all looks good to me, one thing to note, is as mentioned in #76 this is not backwards compatible with users created using SHA256 hashes, so any server owner will need to either manually reset all users passwords or create whole new user accounts. It may be worthwhile adding migration logic. (Check if hash starts with `$argon2id$`, if not, match against SHA256, if that matches then update the stored password to argon2 and sign them in.)

ELLEation:
Indeed it is a breaking change. I'm working on a helper script, but figured it'd be good to get this out there.

diogomartino:
Yea looks good at first glance. Please implement what @TheraNinjaCat suggested so it won't break existing users, I might keep that logic for a few versions and then remove it.

ELLEation:
@TheraNinjaCat @diogomartino Was able to squeeze this in today, I dropped in my current Sharkord instance's database that I'm hosting with a bunch of SHA256 passwords and it worked flawlessly.

apps/server/src/helpers/safe-compare.ts
diogomartino, line 1: Move this to `helpers` instead of utils.

apps/server/src/http/login.ts
```diff
@@ -126,8 +127,33 @@ const loginRouteHandler = async (
...
+  //Logic to handle legacy SHA256 passwords and migrate them to argon2
+  const isPasswordArgon = existingUser.password.startsWith('$argon2');
+  let passwordMatches = false
+
+  if (isPasswordArgon) {
+    passwordMatches = await Bun.password.verify(data.password, existingUser.password);
+  }
+  else {
+
+    console.log("[auth]: SHA256 password hash detected, attempting to upgrade to argon2")
```
diogomartino: Remove console log but maybe switch to logger.debug.

apps/server/src/http/login.ts
```diff
@@ -126,8 +127,33 @@ const loginRouteHandler = async (
...
+    if (passwordMatches) {
+      const argon2Password = await Bun.password.hash(data.password)
+
+      await db
+        .update(users)
+        .set({
+          password: argon2Password
+        })
+        .where(eq(users.id, existingUser.id))
+        .run();
```
diogomartino: `.run()` is not needed

diogomartino (changes requested):
Make those little adjustments and resolve conflicts, then I'll merge. Ty!

apps/server/src/http/login.ts
ELLEation: Switched it to use logger.debug.

Good info to keep around if necessary.

apps/server/src/http/login.ts
ELLEation: Removed

apps/server/src/helpers/safe-compare.ts
ELLEation: Moved

diogomartino:
@ELLEation don't know if u noticed but there's some conflicts. Can't merge without solving them

ELLEation:
Should all be gravy now.

Issue #197, opened 2026-02-14
Compress videos


6004cc0-cmyk:
When a user uploads a video to a chat, compress it so it takes less disk space. I personally use the following to compress my videos locally:

```
ffmpeg -y \
-i <input> \
-map 0:v:0 -map 0:a? \
-vf scale=-2:720 \
-c:v hevc_nvenc \
-preset p6 -tune hq \
-rc vbr -multipass fullres \
-profile:v main -pix_fmt yuv420p \
-spatial_aq 1 -temporal_aq 1 -aq-strength 8 \
-cq 23 \
-c:a copy \
-movflags +faststart \
-progress pipe:1 \
<output>
```

Let the users with the right permission to choose what quality the videos will be compressed to: 720p, 1080p, or 4k.

diogomartino:
Duplicate of #191

Issue #191, opened 2026-02-14
Remove metadata from an image and compress it


6004cc0-cmyk:
This is very important for both privacy and storage!

When a user uploads an image (may it be as an avatar, cover, or attachment to a chat), remove the metadata from the image and also compress the image so it takes less space.

Suggestions: 45% quality and 2000x2000 as size. A file that takes 2,3 MB, can be below 100 kB when compressed with those settings.

Also make an option for the owner to compress all uploaded images that has not been compressed yet.

Aasikki:
This would be nice but I'd also want an option to upload uncompressed media. Maybe default to compressing for the default role and make it possible to grant a permission for roles/users to choose at time of upload if they want to upload it uncompressed (still defaulting to compressing so it's a conscious decision because it won't be needed for everything).

TheraNinjaCat:
I've been looking into changing a few parts on how file uploads are handled, and have put together a system for both compressing arbitrary files using gzip upon upload, as well as re-encoding images and videos as webp and H265 respectively to address both this and #197. It's definitely not ready for a PR yet, still a work in progress, nothing's configurable yet etc, however I'd appreciate feedback on it here: https://github.com/TheraNinjaCat/sharkord/tree/file-upload-changes.

One part I'm not entirely sure how to solve is when doing video compression, as it's all behind awaits the user's upload hangs while it processes the file, if multiple people were to try to upload videos at the same time I'm sure many server's would struggle.

zdlovett:
I'd like to add a +1 for making the compression settings something that are optional and configurable. My friend group has a lot of photographers and being able to share uncompressed images with the metadata is actually a big feature for us! I know thats kinda the opposite of most use cases, but one of the things thats really nice about a self hosted solution like this is I can use my media server so storage is never really an issue.

diogomartino:
Image and video optimization will be possible via plugins, not natively.

PR #279, opened 2026-02-18
File upload changes

TheraNinjaCat:
## Summary

This is a fairly sweeping change to the file upload system, however it fundamentally functions very similarly.

 I have added image and video compression, on top of a general gzip compression to all files. The image and video compression can be configured under storage settings.

This addresses both #191 and #197, although further improvements could be made by exposing hardware encoding options to the server owner, it currently uses software encoding.

The gzip compression is done server-side, however the decompression is done client-side through the use of the `Content-Encoding: gzip` header.

It also moves all server-side file upload management tasks to the file-manager class.

Files are no longer addressed purely by their original (uniquified) file name in `http/public.ts`, instead it is a combination of UUID and original file name. This lowers the guessability of file names, essentially mitigating unauthenticated access to files.

Additionally, all files are now stored in double-nested byte folders based on a portion of the UUID. This helps prevent file-system overhead caused by too many files being stored in a single folder.

File names are now sent between client and server in UTF-8 URI encoding to allow non-ascii file names to be uploaded.

This is unfortunately a breaking change as some database structure changes needed to be made for uploaded files, a migration script could be made, however I have not got one ready.


diogomartino:
Lots of changes so it's gonna take a while to go through them all but from what I've seen in a glance:
- it definitely needs a migration so it doesn't break existing servers
- did you test the compiled version? i doubt gonna carry ffmpeg or sharp correctly for each specific build target
- idk how that ffmpeg lib is implemented but it needs to run it in a different process or in a worker so it doesn't block the main thread

The file upload changes sound good but the image processing it's probably better off in a plugin. It's a nice to have that lots of people will want but I'm trying not to include non critical binaries into the compiled version. Sharp will be a problem as well.

TheraNinjaCat:
Thanks for the quick review, and thoughts.

1. I can start work on migration code, I'm not 100% familiar with the full project (just hyperfixated on a small portion), is there an intended section for migration code, or is an auxiliary shell script sufficient?
2. I didn't check the compiled version, however I am inclined to agree with your hunch here, I also doubt it'll link properly, especially for ARM builds.
3. The `fluent-ffmpeg` library does indeed shell out to a separate process, and my changes put it in an async promise that both updates the db and pushes a message update after it's finished so it doesn't block the main process.
3.1. One thing to note is the fluent-ffmpeg library, while it seems to be the most popular ffmpeg wrapper, [it no longer has any active maintainers](https://github.com/fluent-ffmpeg/node-fluent-ffmpeg?tab=readme-ov-file#fluent-ffmpeg-is-now-deprecated). It still works with the settings I've used here, however it does need to be noted that it may break with ffmpeg updates.
4. Given both 2. and 3.1 above, I think I also agree with your suggestion of media processing being moved to a plugin. I can look into figuring out the plugin system and integrating with that, I suspect I'll need to add a hook in the file manager class for plugin input, but that can be a PR separate to this one if it does need it.

diogomartino:
The SDK will need to be extended yes, I'll see if I can add a hook that runs either in the upload route or in the file manager. Where do you think the hook should be located?

TheraNinjaCat:
Personally, I would expect the hook to be in `file-manager.ts` on line 338, exactly replacing my `reEncodeMedia` call. This is essentially the start of a promise chain that aims to process and compress the file asynchronously and update the database with the on-disk size to let people store more files.

It would allow users to hook into the tempFile and make any changes they like to the uploaded file, before the blanket gzip compression is attempted, while still allowing everything to happen asynchronously without interrupting the message send process if their processing takes a long time (such as what happens with the video and gif compression with ffmpeg and sharp).

I'm also just thinking, potentially the blanket gzip compression should be exposed as a storage setting to administrators as well, as while I didn't notice any performance impact for files I tried uploading, some lower-powered systems may not find the CPU-Storage tradeoff worth it and not desire the behaviour.

diogomartino:
Yes, making you able to configure if you want gzip compression is good. But maybe I would add it to the `config.ini` since it's such a "low level" option...

I'll investigate the best way to add the hook, but yes that's probably the best spot. You would do something like:

```
  ctx.events.on("file:uploaded", async ({ tempFile }) => {
    // do things with the uploaded file
  });
```

TheraNinjaCat:
That more or less sounds good to me, as long as the plugin system can hook into the promise chain nicely so things are done in order. I can't work on this right this second, but I'll remove the file encoding stuff from this PR and look at fixing up the failing tests to work with the new `file-manager` interface correctly.

TheraNinjaCat:
I've fixed up all the tests and a few other things along the way. The only outstanding thing is complete migration, for which I'm unsure about how best to implement it. Things that need to be changed to migrate are the database needs updating, all existing files will need to have their name changed to a uuid, and to be moved into the correct new folder in public/??/??/uuid. Optionally do a 1-pass gzip across all the files and update the database's compressed boolean to match.

diogomartino:
That will be tricky. I don't think drizzle supports migrations via scripts, so from the migration you can't move files around. Idk if this is the best time to make such a drastic change tbh.

rufflepot:
I really like the job you did on this one @TheraNinjaCat. File upload as a bit lacking right now and your solution is beautiful. While breaking changes are always sad, implementing this later would be even worse. On the other hand, there might be more breaking changes coming eventually and collecting a bunch of them to implement at once might be preferable. I for one wouldn't be too bothered by alpha software straight from the dev pipeline introducing breaking changes every once in a while. The time of stability is later - this is the age of CHAOS!

blindmikey:
Alternatively it would be great if this new method applies to all media uploaded after the update while being backwards compatible for all media already uploaded.

diogomartino:
@Blindmikey yes, that's probably what's gonna happen. Just not sure when, there are more important things being done now.

diogomartino:
I'm gonna close this as stale. Your work is not gonna go to the trash tho, when I pick this up again sometime in the future I'll use some of your code and credit you there. This was just unfortunate timing. Sorry about that.

PR #290, opened 2026-02-19
Feature (#217) - Added ability to have locally stored draft messages on text channels.

### LINKED ISSUE #217: Request: Save draft of typed message when switching chats
If you start typing something in one chat then switch to another, what was typed is delete.
I suggest adding a way to say a draft of what was typed so when you go back to the chat it is still there
or have a message asking for confirmation before switching chat channels.

toby-short:
## Summary
Added the ability to have draft messages saved locally for the user, so they don't just disappear when you change channels. These are stored locally in local storage but are keyed on channel and user ID.

Closes #217

### DIFF
--- a/apps/client/src/components/channel-view/text/index.tsx
+++ b/apps/client/src/components/channel-view/text/index.tsx
@@ -167,7 +167,7 @@ const TextChannel = memo(({ channelId }: TChannelProps) => {
     setNewMessage(value);
 
     setDraftMessage(draftChannelKey, value);
-  }, []);
+  }, [setNewMessage, setDraftMessage]);
 
   if (!channelCan(ChannelPermission.VIEW_CHANNEL) || loading) {
     return <TextSkeleton />;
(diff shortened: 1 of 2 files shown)


apps/client/src/components/channel-view/text/index.tsx
```diff
@@ -160,6 +163,12 @@ const TextChannel = memo(({ channelId }: TChannelProps) => {
     [removeFile]
   );
 
+  useEffect(() => {
+    if (!newMessage) return;
+
+    setDraftMessage(draftChannelKey, newMessage);
+  }, [newMessage]);
+
```
diogomartino: Bad practice, shouldn't use useEffect to do this. Add a new callback like:

```
  const setNewMessageHandler = useCallback((value: string) => {
    setNewMessage(value);

    // set draft here
  }, []);
```

and then use the `setNewMessageHandler` where `setNewMessage` was being used

toby-short: Sorting it now :)

toby-short: Done.

PR #302, opened 2026-02-19
feat(273) - Hide user Identity/IP/Location in "Details" by default

### LINKED ISSUE #273: Request: Hide user Identity/IP/Location in "Details" by default
Image

As a user with the right permissions, you can open up "Moderation view" on a given user and inside is the "Details" section. Inside the "Details" the users Identity, IP Address and Location are shown in clear text.

I suggest that these 3 details be hidden by default, with something like an eye icon next to them allowing the user to uncover them if necessary.

Why?
To make it a bit less easy to leak someones Identity/IP/Location by accident, while for example screen sharing.

Pesuri:
## Summary

Closes #273
---

Hide user identity, IP and location in Details by default.
Allow viewing these with correct permission by clicking the eye icon in the row.
Role needs additionally "View user identity, IP address and location" permission to be able to see these. Without permission these rows are not shown.

## Additional Context

kuva
kuva
<hr/>
No permission to view identity, IP address and location.<br/>
kuva

### DIFF
--- a/apps/client/src/components/mod-view-sheet/details/index.tsx
+++ b/apps/client/src/components/mod-view-sheet/details/index.tsx
@@ -38,7 +38,7 @@ const Row = memo(({ icon, label, value, details, hidden = false }: TRowProps) =>
     </span>
   );
   
-  let [visible, setVisible] = useState(!hidden);
+  const [visible, setVisible] = useState(!hidden);
 
   if (details) {
     valContent = <Tooltip content={details}>{valContent}</Tooltip>;


apps/client/src/components/mod-view-sheet/details/index.tsx
```diff
@@ -10,27 +10,38 @@ import {
...
+const Row = memo(({ icon, label, value, details, hidden = false }: TRowProps) => {
   let valContent = (
     <span className="text-sm text-muted-foreground truncate max-w-[160px]">
       {value}
     </span>
   );
+  
+  const [visible, setVisible] = useState(!hidden);
+  
+  function handleEyeClick() {
```
diogomartino: Always use ES6 syntax to keep consistency. Also this codebase uses a always memo approach. In this case  you don't need it tho, you can use the `setVisible(!visible)` directly instead of creating a new handler.

apps/client/src/components/mod-view-sheet/details/index.tsx
```diff
@@ -42,7 +53,25 @@ const Row = memo(({ icon, label, value, details }: TRowProps) => {
...
-      {valContent}
+      {visible ? valContent : "***"}
+      {visible && hidden &&
+        <span
+          role="button"
+          onClick={handleEyeClick}
+          className="text-muted-foreground inline-flex h-6 w-6 items-center justify-center rounded bg-transparent hover:bg-accent hover:text-foreground cursor-pointer transition-colors focus:outline-none"
+        >
+          <Eye className="h-4 w-4" />
+        </span>
```
diogomartino: This should be an `IconButton` component.

packages/shared/src/statics/permissions.ts
```diff
@@ -18,7 +18,8 @@ export enum Permission {
   MANAGE_INVITES = 'MANAGE_INVITES',
   MANAGE_UPDATES = 'MANAGE_UPDATES',
   MANAGE_PLUGINS = 'MANAGE_PLUGINS',
-  EXECUTE_PLUGIN_COMMANDS = 'EXECUTE_PLUGIN_COMMANDS'
+  EXECUTE_PLUGIN_COMMANDS = 'EXECUTE_PLUGIN_COMMANDS',
+  VIEW_IDENTITY_IP_LOCATION = 'VIEW_IDENTITY_IP_LOCATION'
```
diogomartino: Maybe this is too explicit. Maybe "VIEW_USER_SENSITIVE_DATA" is better?

Pesuri: Seems like a good idea.

diogomartino:
Lint is failing, fix that and I'll merge.

Pesuri:
Fixed

diogomartino (approved):
I noticed now that this is not done all the way because even if you don't have permission you will still receive that data from the server. I'll merge it and fix it myself, ty 👍

Pesuri:
I see. I must take a note about that and remember it in similar cases.

Issue #47, opened 2026-02-10
CTRL + R and general reload removes active session


schemen:
When refreshing the site it seems to kill the active session and require a log-in again.

I'm not sure if this is intended behavior, but my expectation would be it'll simply refresh the site, keeping you logged in.

diogomartino:
It's intended but probably it's a good idea to let the user decide if he wants to auto login or not.

BabaBooey84:
Thanks for clarifying that this is intended behavior for now - I was scratching my head a bit 🙂

Just wanted to add some user feedback: not having login persist across a page refresh can be quite confusing for end users, especially those who aren’t technical. A refresh (or accidental reload) feels like a very normal action in a web app, and being silently logged out makes it seem like something is broken or misconfigured.

A persisted token option, or at least a clearer UX hint that sessions are temporary, would go a long way in improving usability.

Really like the project so far - keep up the great work!

schemen:
Thanks for the clairifcation! And good write up @BabaBooey84, I agree with you.

ELLEation:
I am working on a branch that will utilize session cookies and so far it is looking good.

PR #306, opened 2026-02-20
feat(#47): initial implementation of session cookies

ELLEation:
## Summary

Initial implementation of session cookies for a more persistent login experience. It isn't perfect but a solid start.

Closes #
https://github.com/Sharkord/sharkord/issues/47

---

## Additional Context

Further down the road I hope to implement rotating session keys to persist logins if remember me is selected. 
This should be a non-breaking change for existing users. I have tested it on my live instance.
CSRF may be a potential risk to consider with this but is arguably better than cached credentials in local storage.


diogomartino:
That probably works but I don't want to use cookies unless it's absolutely necessary. I'll merge a different approach. Ty tho :heart:

PR #309, opened 2026-02-20
fix(#308): Explicitly set password blank in get-user-info.ts

### LINKED ISSUE #308: [Bug]: getUserInfoRoute includes password hashes in response data
### Sharkord Version

0.0.4

### Environment

All

### Steps to Reproduce

The user object from here includes the password field from the database, so any user with `Permission.MANAGE_USERS` is able to retrieve user password hashes.

https://github.com/Sharkord/sharkord/blob/a4a4431aba27252a15c6d309abfe1704ebc4c6d9/apps/server/src/routers/users/get-user-info.ts#L19
https://github.com/Sharkord/sharkord/blob/a4a4431aba27252a15c6d309abfe1704ebc4c6d9/apps/server/src/routers/users/get-user-info.ts#L32
https://github.com/Sharkord/sharkord/blob/a4a4431aba27252a15c6d309abfe1704ebc4c6d9/apps/server/src/db/queries/users.ts#L191-L205

### Expected Behavior

`getUserInfo` should first pass the user object through `clearFields(users, ['identity', 'password'])`  as in `getUsers` before including it in the return data to ensure that at least the password, but ideally both the identity and password hash are not included in server responses.

### Logs / Screenshots

client to server
```
[
    {
        "id": 41,
        "method": "query",
        "params": {
            "input": {
                "userId": 2
            },
            "path": "users.getInfo"
        }
    },
    {
        "id": 42,
        "method": "query",
        "params": {
            "input": {
                "userId": 2
            },
            "path": "users.getInfo"
        }
    }
]
```

server to client
```
{
    "id": 41,
    "result": {
        "type": "data",
        "data": {
            "user": {
                "id": 2,
                "identity": "Username",
                "name": "SharkordUser",
                "avatarId": null,
                "bannerId": null,
                "bio": null,
                "password": "9f86d081884c7d659a2feaa0c55ad015a3bf4f1b2b0b822cd15d6c15b0f00a08",
                "bannerColor": null,
                "createdAt": 1771574992299,
                "updatedAt": null,
                "lastLoginAt": 1771574995760,
                "banned": false,
                "banReason": null,
                "bannedAt": null,
                "avatar": null,
                "banner": null,
                "roleIds": [
                    1,
                    2
                ]
            },
            "logins": [],
            "files": [],
            "messages": []
        }
    }
}
```

TheraNinjaCat:
## Summary

Explicitly overwrites the password field to prevent it being sent to clients.

Closes #308 

---

### DIFF
--- a/apps/server/src/routers/users/get-user-info.ts
+++ b/apps/server/src/routers/users/get-user-info.ts
@@ -6,6 +6,7 @@ import { getMessagesByUserId } from '../../db/queries/messages';
 import { getUserById } from '../../db/queries/users';
 import { invariant } from '../../utils/invariant';
 import { protectedProcedure } from '../../utils/trpc';
+import { clearFields } from '../../helpers/clear-fields';
 
 const getUserInfoRoute = protectedProcedure
   .input(
@@ -23,15 +24,13 @@ const getUserInfoRoute = protectedProcedure
       message: 'User not found'
     });
 
-    user.password = '';
-
     const [logins, files, messages] = await Promise.all([
       getLastLogins(user.id, 6),
       getFilesByUserId(user.id),
       getMessagesByUserId(user.id)
     ]);
 
-    return { user, logins, files, messages };
+    return { user: clearFields(user, ['password']), logins, files, messages };
   });
 
 export { getUserInfoRoute };


TheraNinjaCat:
Notably, this does not remove identity as raised in the original issue, but that is included in the UI so is likely intentionally here.

apps/server/src/routers/users/get-user-info.ts
```diff
@@ -23,6 +23,8 @@ const getUserInfoRoute = protectedProcedure
       message: 'User not found'
     });
 
+    user.password = '';
```
diogomartino: Use the `clearFields` helper so it's easier to track where we are hiding certain fields.

TheraNinjaCat: Done, sorry for not using it initially, I didn't see it also accepted a single object, not just a list, so I didn't think it'd work properly and wanted the fix asap.

PR #315, opened 2026-02-20
fix(#304) Image file extension from uppercase to lowercase

### LINKED ISSUE #304: [Bug]: Image files with uppercased file extension don't show image.
### Sharkord Version

v0.0.4

### Environment

Arch Linux 6.18.9

### Steps to Reproduce

When an image is added to chat and image file's extension in filename is in uppercase, preview image won't be shown. If extension is changed to lower case, preview image works.

### Expected Behavior

File's type being written in uppercase or lowercase shouldn't affect previewing image.

### Logs / Screenshots

First image shows preview, but second doesn't.
Image

Pesuri:
## Summary

Closes #304

---
Well, this was a easier fix than I thought.

## Additional Context
With this fix image file's extensions case doesn't affect if image is shown. Also save file extension name in lowercase into database.
kuva

### DIFF
--- a/apps/server/src/utils/file-manager.ts
+++ b/apps/server/src/utils/file-manager.ts
@@ -236,7 +236,7 @@ class FileManager {
       .insert(files)
       .values({
         name: fileName,
-        extension: tempFile.extension,
+        extension: tempFile.extension.toLowerCase(),
         md5: tempFile.md5,
         size: tempFile.size,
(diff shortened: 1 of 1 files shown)


Pesuri:
That sounds like a good idea.

Pesuri:
Now it should save extension name in lowercase.

PR #346, opened 2026-02-22
fix(345): exclude deleted users from count

### LINKED ISSUE #345: [Bug]: Deleted default SharkordUser counts to total members count
### Sharkord Version

0.0.5

### Environment

Development

### Steps to Reproduce

1. Create a user account
2. Use token to acquire owner rights
3. Delete default Sharkord user

The deleted default Sharkord user will still be included in the member count:

Image

### Expected Behavior

The member count should show the actual number of members

### Logs / Screenshots

Image

diogomartino:
## Summary

Closes #345

### DIFF
--- a/apps/client/src/components/right-sidebar/index.tsx
+++ b/apps/client/src/components/right-sidebar/index.tsx
@@ -55,6 +55,19 @@ const RightSidebar = memo(
 
     const hasHiddenUsers = users.length > MAX_USERS_TO_SHOW;
 
+    const usersCount = useMemo(
+      () =>
+        users.reduce((count, user) => {
+          // exclude the special "Deleted User" from the count if it exists
+          if (user.name !== DELETED_USER_IDENTITY_AND_NAME) {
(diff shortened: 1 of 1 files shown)


abynox:
Hey @diogomartino , thanks for the fast fix :D

I was just about to open a PR for the issue and I think that my solution might be the better, since it would avoid iterating over users multiple times
What do you think?

```typescript
const { usersToShow, filteredCount } = useMemo(() => {
  const filtered = users.filter(
    (user) => user.name !== DELETED_USER_IDENTITY_AND_NAME
  );

  return {
    usersToShow: filtered.slice(0, MAX_USERS_TO_SHOW),
    filteredCount: filtered.length,
  };
}, [users]);
```

diogomartino:
@abynox yea you are right, your solution is a lot better. I reopened the issue so it's easier to track, can you open a pr? Ty!

abynox:
Sure thing

Issue #73, opened 2026-02-11
Request: Create User / Sign-in Improvement


ZandaBear:
Two primary things that not only would improve user experience but also reduce the amount of accidental duplicated accounts.

1. "Create User" screen that is separate from login. Bonus for ToS link as well.
2. Remove case sensitivity from Username field. ABC123, AbC123, and aBc123 for example should be one account, not three.

Aasikki:
Also to add, the create user screen should also ask for a username (the one shown to everyone) and optionally a profile picture.
Setting a username should be mandatory but profile pic of course not. Currently I have some people who don't bother to set a username, which makes it hard to identify who should be banned/kicked if I ever find myself in a situation where I need to do so :)

ghost:
> Also to add, the create user screen should also ask for a username (the one shown to everyone) and optionally a profile picture. Setting a username should be mandatory but profile pic of course not. Currently I have some people who don't bother to set a username, which makes it hard to identify who should be banned/kicked if I ever find myself in a situation where I need to do so :)

Yeah issue #59 addresses that.

diogomartino:
Will be implemented in #59

PR #348, opened 2026-02-22
Feature: Split login and registration pages

### LINKED ISSUE #59: Add initial profile setup prompt on first login
The current flow of getting a default display name until changed is mostly fine, but can become a bit messy when people get in the server unattended and forget to change their names. I think it would be nice to show new users a window where they would be able to select a name and, optionally, an avatar before continuing.

TheraNinjaCat:
## Summary

Splits the login and registration functions into 2 screens and adds a field for Display Name during registration.

Splitting them up was discussed in #73 to prevent duplicate users being created by accident, and adding a field for Display Name during registration was discussed in #59. This addresses both issues.

Closes #59

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,30 @@
+name: E2E Tests
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
+jobs:
+  ci:
+    name: Run E2E Tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Bun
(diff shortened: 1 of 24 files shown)


diogomartino:
Only having a single form for login and registration was intended and it's not gonna change, at least not for now. That ticket talks about sort of an "onboarding" experience when you first join the server, not having separate experiences for login and registration.

So the idea is to when a user logs in for the first time, a new modal shows up where the user can setup his profile. This can be a shared component that's exactly the same as we have on user settings. This modal also can be extended via plugins.

I advise discussing features first before implementation to avoid this situation, which sucks for everyone involved. Please continue the discussion in #59 if you want.

PR #349, opened 2026-02-22
feat(#337): Show user banners in voice chat

### LINKED ISSUE #337: [Feature]: Show a user's banner in the background while in voice chat
### Problem

A users banner is only shown in their popover and is not used anywhere else.
More personalization is always better :)

### Proposed Solution

Use the banner as the background for voice cards.

Something like this:
Image

LMK if this feature is wanted or if you have any thoughts.

Herbycheck:
## Summary
Closes #337

The banner of a user is now used as the background for their voice card when inside a call. This can be toggled via the voice settings.

## Additional Context

If a user doesn't have a banner set, the default background is shown instead

 

Toggle in voice settings


apps/client/src/components/channel-view/voice/voice-user-card.tsx
```diff
@@ -66,7 +69,16 @@ const VoiceUserCard = memo(
...
+        {voiceUser.banner && showUserBanners && (
+          <div
+            className="h-full w-full rounded-t-md bg-cover bg-center blur-sm brightness-50 bg-no-repeat absolute inset-0"
+            style={{
+              backgroundImage: `url(${getFileUrl(voiceUser.banner)})`
+            }}
+          />
+        )}
+
+        {(!voiceUser.banner || !showUserBanners) && <CardGradient />}
```
diogomartino, line 81: A ternary would be cleaner here

Issue #359, opened 2026-02-23
[Bug]: Docker image is running as root inside the container


TheraNinjaCat:
### Sharkord Version

0.0.5

### Environment

Docker

### Steps to Reproduce

1. Run the latest sharkord docker container
2. Run `docker exec -it sharkord /bin/bash` in a separate terminal
3. Run `apt update &>/dev/null; apt install procps -y &>/dev/null && ps -e -o pid,comm,user,uid,group,gid | grep sharkord` in the new shell
4. Observe the output `1 sharkord        root         0 root         0` showing the process running as root

### Expected Behavior

The output should be something like ` 2986 sharkord        bun       1000 bun       1000` showing the sharkord process running as a non-root user.

killfrenzy96:
You can currently run the container without root like this:
```
mkdir -p ./data
chown 12345:12345 ./data
docker run \
  --user 12345:12345 \
  -e HOME=/home/bun \
  -p 4991:4991/tcp \
  -p 40000:40000/tcp \
  -p 40000:40000/udp \
  -v ./data:/home/bun \
  --name sharkord \
  sharkord/sharkord:latest
```
Or for a more hardened config:
```
mkdir -p ./data
chown 12345:12345 ./data
docker run \
  --user 12345:12345 \
  -e HOME=/home/bun \
  --security-opt no-new-privileges \
  --read-only \
  --cap-drop all \
  -p 4991:4991/tcp \
  -p 40000:40000/tcp \
  -p 40000:40000/udp \
  -v ./data:/home/bun \
  --name sharkord \
  sharkord/sharkord:latest
```

Bobbysx:
This is my config without Root:

```
services:
  sharkord:
    image: sharkord/sharkord:latest
    container_name: sharkord
    user: 1000:1000
    environment:
      - XDG_CONFIG_HOME=/config
    ports:
      - 4991:4991/tcp
      - 40000:40000/tcp
      - 40000:40000/udp
    volumes:
      - /opt/sharkord:/config/sharkord
    restart: unless-stopped
    cap_drop:
      - ALL
    cap_add:
      - NET_BIND_SERVICE
    security_opt:
      - no-new-privileges:true
networks: {}
```

PR #360, opened 2026-02-23
Feature: Docker container now installs optional dependencies

### LINKED ISSUE #107: Docker container requires python3 for Sharkord Music Bot plugin
The docker container doesn't have python installed so it is[ required to add python3](https://github.com/yt-dlp/yt-dlp?tab=readme-ov-file#recommended) after if you want to use the [Sharkord Music Bot](https://github.com/diogomartino/sharkord-music-bot) plugin.

TheraNinjaCat:
## Summary

Added init.sh that installs dependencies from the SHARKORD_DEPENDENCIES environment variable.
It then launches sharkord as the bun user (1000:1000) to prevent running the app as root.

Closes #107 and #359

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ ENV RUNNING_IN_DOCKER=true
 
 COPY apps/server/build/out/sharkord-linux-x64 /tmp/sharkord-linux-x64
 COPY apps/server/build/out/sharkord-linux-arm64 /tmp/sharkord-linux-arm64
+COPY init.sh /init.sh
 
 RUN set -eux; \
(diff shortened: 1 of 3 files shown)


diogomartino:
I don't like this too much. At least for now, I prefer to keep things simple. If you need extra dependencies, just take the base image and extend it to create your own. Eg: if you need python, this:

```
FROM sharkord/sharkord:latest

RUN apt-get update && apt-get install -y python3
```

should work, I think.

The root user change is needed tho, but I'll add it in some other PR to keep things clean.

PR #392, opened 2026-02-25
feat(80): Add noise gate + audio level performance improvements

### LINKED ISSUE #80: Feature request: Voice activity detection

schemen:
## Summary
This PR implements an audio gate using browser native AudioWorklet API as audio processor. All browsers are supported (https://developer.mozilla.org/en-US/docs/Web/API/AudioWorklet#browser_compatibility) but just in case, the fallback is a failsafe, and it just won't turn on the feature.

We can probably argue about the defaults, currently:
* Defaults to ON at -48 dB

Closes #80 

## Additional Context

Since the audio level display is my previous crime, I've spent some time improving the performance and noticed that on each update the entire settings page gets rendered. I've separated that now and thus, it performs now really well at high refresh rate. It now renders at the browser speed and the data gets refreshed all 20ms which should be enough.

Additionally, to make finding easier, added a peak meter to it so you can easily set the gate where the peaks are.

image

image


diogomartino:
Nice! Did you check if it works in all browsers? Last time I tried audio worklets firefox had some strange behaviors...

schemen:
> Nice! Did you check if it works in all browsers? Last time I tried audio worklets firefox had some strange behaviors...

I've relied on https://developer.mozilla.org/en-US/docs/Web/API/AudioWorklet#browser_compatibility to be honest! I do use Firefox and tested a bit on Chrome, others no, I've not tested any other ones.

Though, there are fallbacks =) Not for the gate though, it would just be disabled.


EDIT:

I also tested on my iPhone IOS Safari where it works 👍

diogomartino:
Nice. Gonna take a look at this probably tomorrow.

schemen:
Hm, I think lowering the default to maybe -50 or -55dB wouldn't be bad. I can't say - I just have my own data so far. 

And/or maybe we leave it as opt in? Any opinions?

SSTHacksword:
I personally feel features like this are best if they are on by default. Gating is something that helps clear up the voice channels and I am personally really excited to see this implemented. Less tech savvy users will most likely not even look for this unless they are told to.

diogomartino:
I've been testing it and it looks good. I think it should default to off since the default values are completely arbitrary and it will really depend on the mic and ambient where the user is in, so in the end the user should tweak it himself. I'm not entirely sure if the audo worklets are gonna just work in the compiled version, lets hope so. Gonna merge.

PR #408, opened 2026-02-27
feat(#389): mark edited messages

### LINKED ISSUE #389: [Feature]: Edited messages should be marked
### Problem

There should be some kind of marker to indicate that a message has been edited.
Perhaps it should also show when it was edited and by whom.
Users with "manage messages" permissions could abuse the message editing feature without anyone noticing.

### Proposed Solution

Show a small “Edited by user” tag on edited messages. When hovering over it, the time at which the message was edited could be displayed

0x6DD8:
Closes #389 

image


diogomartino (approved):
LGTM

PR #426, opened 2026-02-28
feat(258): add PWA manifest for installable web app support

### LINKED ISSUE #258: [Feature]: PWA - Progressive Web App - Allow users to "install" as a web app.
### Problem

Currently users must open their browser and navigate to a url. With a PWA manifest file, users could open a sharkord server as they would any other computer application, without any need to wrap sharkord in something like electron.

### Proposed Solution

Add a manifest.json file exposed to the public would allow browsers such as chrome to prompt the user to "install as web app" which would pin the web application to their desktop/task bar so that they can open it in a dedicated window akin to a normal application. ( github.com does this! )

https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Manifest#example_manifest

Doesn't have to be complex, and can be filled out with the server's settings for server name, image, etc. 

Image

nylesninetynine:
Closes #258

Adds a `manifest.json` and PNG icons to enable PWA installability in supported browsers (Chrome, Chromium, Edge), allowing users to install Sharkord as a standalone desktop app without needing Electron or any wrapper.

Changes:
- Added `apps/client/manifest.json`
- Added `apps/client/public/icon-192.png` and `icon-512.png` converted from existing `logo.webp`
- Added `<link rel="manifest">` to `index.html`


diogomartino (approved):
LGTM

jakealstad:
oh awesome, when's the next release?? :)

PR #429, opened 2026-03-01
feat(#401) Add basic text format tools

### LINKED ISSUE #401: [Feature]: Basic Markdown
### Problem

Currently text is plain-text only (sorry if I am wrong on this)

### Proposed Solution

Add some basic markdown functionality like underlining, bold, italicizing, code, lists, etc.

Pesuri:
## Summary

Closes #401.
Adds basic text formatting tools. Bold, italic, underline, strike, link, code, codeblock, horizontal line and blockquete.

Sharkord text editor already supports basic markdown inputs. Editor changes markdown inputs into HTML tags, but sanitization didn't allow all tags and CSS was missing configurations. These changes allow markdowns also work better.

At the moment hyperlink and underline are identical. Should we change link color to something else?
It might be a good idea to add lists also.

This would also change message sending from enter to ctrl+enter. Linebreak with shift+enter didn't work right with codeblock. This might need more work. Or is it better this way? It is way too easy to send message by pressing only enter but that is only my opinion.

## Additional Context

kuva

### DIFF
--- a/apps/client/src/components/channel-view/text/renderer/index.tsx
+++ b/apps/client/src/components/channel-view/text/renderer/index.tsx
@@ -97,7 +97,7 @@ const MessageRenderer = memo(
           )}
         >
           {messageHtml}
-          {message.editedAt && (
+          {message.editedAt ? (
             <Tooltip
               content={
                 <div className="flex flex-col gap-1">
@@ -118,7 +118,7 @@ const MessageRenderer = memo(
                 (edited)
               </span>
             </Tooltip>
-          )}
+          ) : (<span></span>)}
         </div>
 
         {allMedia.map((media, index) => {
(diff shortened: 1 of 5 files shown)


apps/client/src/components/channel-view/text/renderer/index.tsx
```diff
@@ -118,6 +118,8 @@ const MessageRenderer = memo(
                 (edited)
               </span>
             </Tooltip>
+          ) : (
+            <span></span>
```
Pesuri, line 122: CSS formatting doesn't work well without this for some reason.

killfrenzy96:
I feel like implementing a markdown renderer library should be better. This would put it in line with Discord's formatting capabilities. Markdown is a standard used by many applications and websites.

Pesuri:
Sharkord editor seems to already support basic markdowns.
It is missing CSS configurations that are added in this PR.
Edit. and sanitization is missing allowed tags.

diogomartino:
Gonna close this one. If I'll add something like this it will be on #541

Issue #278, opened 2026-02-18
[Feature]: Hide category if user can't see any channels within


Blindmikey:
### Problem

If I set permissions based-viewing of all channels within a category, users who don't have those permissions just see an empty category.

### Proposed Solution

If a user can't see any channels within a category, hide or do not render that category to that user.

diogomartino:
Agreed. Hide categories if:
-  doesn't have any channels in it (no permissions or literally no channels)
- AND 
- user doesn't have MANAGE_CATEGORIES permission

dankreefer:
Can it be taken a step further and assign permissions to the category itself and issue as Parent/child permissions.

PR #455, opened 2026-03-02
feat(278): Hides categories when users can't see any channels within them

StephenArg:
## Summary
If admin users set private permissions based-viewing of all channels within a category, users who don't have permissions for any of the channels within a category no longer see the empty category section. Also works if the category section is empty.

Added a new hook to check if user has access to any of the channels in a list of channel ids (taken from each category)

Closes #
#278


diogomartino:
Check what you commit before opening a PR. You committed a bunch of test-data files. To run tests you should do `bun run test`, not `bun test` otherwise that happens.

StephenArg:
> Check what you commit before opening a PR. You committed a bunch of test-data files. To run tests you should do `bun run test`, not `bun test` otherwise that happens.

Ah that's where that came from. Ok I'll remove it now

apps/client/src/components/left-sidebar/categories.tsx
```diff
@@ -35,6 +36,11 @@ type TCategoryProps = {
 };
 
 const Category = memo(({ categoryId }: TCategoryProps) => {
+  const can = useCan();
+  const channels = useChannelsByCategoryId(categoryId);
+  const channelCan = useChannelInCategoryCan(
+    channels.map((channel) => channel.id)
+  );
```
diogomartino: This map here will make the hook recalculate every render. This data should come from a selector. Also the name `useChannelInCategoryCan` doesn't make too much sense. It should be called `useHasVisibleChannelsInCategory` which accepts the category id and inside you should cache the derived state as much as possible.

apps/client/src/features/server/hooks.ts
```diff
@@ -97,6 +101,30 @@ export const useChannelCan = (channelId: number | undefined) => {
   return can;
 };
 
+export const useChannelInCategoryCan = (channelIds: number[]) => {
+  // Returns true if the user can view at least one of the channels in the category
```
diogomartino: If you gonna comment a what a function does, comment above the function so IDEs can pick it up and show the comment on auto completes.

StephenArg: Fixed

apps/client/src/components/left-sidebar/categories.tsx
StephenArg: Fixed

Issue #388, opened 2026-02-25
[Bug]: Auto Login not working in Firefox


EricC549:
### Sharkord Version

v0.0.7

### Environment

Using Windows to access site / Sharkord hosted on Ubuntu Server

### Steps to Reproduce

Go to self hosted site in Firefox such as sharkord.example.com

At login screen check Login automatically

Enter username and password

Refresh page 

Not automatically logged in

Works fine in Edge and Chrome

### Expected Behavior

Expected to be automatically logged in on page refresh

### Logs / Screenshots

<img width="585" height="284" alt="Image" src="https://github.com/user-attachments/assets/8f7ea294-717d-4131-a029-26e495905a3e" />

<img width="557" height="271" alt="Image" src="https://github.com/user-attachments/assets/0959eec5-07da-4410-b528-8ae3c060ad85" />

EricC549:
Also tried on a separate device that has never accessed the site in Firefox and it does not work there either

ComputerElite:
Same here. The issue seems to be related to the cleanup function removing the `auto-login-token` and `auto-login` entries from localstorage. Uncommenting these lines does fix the issue: https://github.com/Sharkord/sharkord/blob/8ed5066f4b2bba4476d1017ba6de2f49217a66d3/apps/client/src/lib/trpc.ts#L79-L80

EricC549:
Thank you for that info! Unfortunately, I am currently running Sharkord in k3s - so worst case I can fork the project with that fix. But I am sure this will be fixed officially at some point.

diogomartino:
Does it never work for you? Or just sometimes? I installed Firefox here and tried a couple of times and it worked.

EricC549:
It never works for me in Firefox on any device. Chrome and Edge work fine. I don’t know if it helps but I’m using Caddy as the reverse proxy.

pretzelise:
It's consistently kicking me back to the login screen in Firefox too. Also using caddy for reverse proxy.

diogomartino:
Kicking you back even without refreshing?

pretzelise:
No, it's only consistently happening on reload.

Aasikki:
For me, if I just close the tab and open sharkord in a new tab, auto login works fine, but if I refresh the page, it kicks me out.

pretzelise:
> For me, if I just close the tab and open sharkord in a new tab, auto login works fine, but if I refresh the page, it kicks me out.

[...]

PR #459, opened 2026-03-03
fix(388): found and corrected auto-login issue for firefox

StephenArg:
## Summary
Fixed issue causing auto-login to not work on firefox.

It wasn't setting the localstorage key correctly and also firefox was triggering the cleanup function when refreshing while chrome does not. That function erased the auto-login token in localstorage. Now it shouldn't activate when refreshing.

Closes #
#388

### DIFF
--- a/apps/client/src/lib/trpc.ts
+++ b/apps/client/src/lib/trpc.ts
@@ -7,8 +7,7 @@ import {
   LocalStorageKey,
   removeLocalStorageItem,
   removeSessionStorageItem,
-  SessionStorageKey,
-  setLocalStorageItemBool
+  SessionStorageKey
 } from '@/helpers/storage';
 import { type AppRouter, type TConnectionParams } from '@sharkord/shared';
 import { createTRPCProxyClient, createWSClient, wsLink } from '@trpc/client';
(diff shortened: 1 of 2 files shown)


apps/client/src/lib/trpc.ts
```diff
@@ -18,6 +18,15 @@ let trpc: ReturnType<typeof createTRPCProxyClient<AppRouter>> | null = null;
 let currentHost: string | null = null;
 let isCleaningUp = false;
 
+// Firefox fires WebSocket onClose during page refresh; Chrome does not. When navigating away,
+// we must not clear auto-login localStorage or it will be lost on refresh in Firefox.
+let isNavigatingAway = false;
+if (typeof window !== 'undefined') {
```
diogomartino: This check is unnecessary, window will always be available in this context

apps/client/src/lib/trpc.ts
```diff
@@ -87,9 +96,12 @@ const cleanup = () => {
...
 
   // cleanup can be called due to various reasons (manual disconnect, connection error, auto-login failure, etc).
-  // so we remove any persisted auto-login token to prevent auto-login loops
-  removeLocalStorageItem(LocalStorageKey.AUTO_LOGIN_TOKEN);
-  setLocalStorageItemBool(LocalStorageKey.AUTO_LOGIN, false);
+  // so we remove any persisted auto-login token to prevent auto-login loops.
+  // skip this when navigating away (refresh/close) - Firefox fires onClose during refresh, Chrome does not
+  if (!isNavigatingAway) {
+    removeLocalStorageItem(LocalStorageKey.AUTO_LOGIN_TOKEN);
+    setLocalStorageItemBool(LocalStorageKey.AUTO_LOGIN, false);
```
diogomartino: Also remove the `setLocalStorageItemBool(LocalStorageKey.AUTO_LOGIN, false);`, was wrong before, no need to actually remove the auto login toggle, the token is enough.

apps/client/src/screens/connect/index.tsx
```diff
@@ -170,6 +170,10 @@ const Connect = memo(() => {
             data-testid={TestId.CONNECT_AUTO_LOGIN_SWITCH}
             onClick={() => {
               onChange('autoLogin', !values.autoLogin);
+              setLocalStorageItemBool(
```
diogomartino: This is not needed here, we only persist values if the login is successful.

StephenArg:
> Just make these small changes please. But I think this should fix it.

Done 👍

Issue #263, opened 2026-02-17
[Feature]: Improve voice channel noise suppression


Giggitybyte:
### Problem

When enabled, the current noise suppression works well enough to filter out sounds that are more than a few feet away from me. However, it still allows for non-human audio (key strokes and mouse clicks) to come through loud and clear. Ambient activity is not filtered out by the noise suppression.

This is an annoyance when playing video games, or just eating food while chatting.

### Proposed Solution

Instead of using the browser's WebRTC noise suppression, implement a more robust noise suppression library on the client side. 

A few I've found are:
https://github.com/hayatialikeles/dtln-rs
https://github.com/sapphi-red/web-noise-suppressor

Giggitybyte:
Whoops, just realized this duplicates #84. My bad.

PR #460, opened 2026-03-03
feat: incorporate dtln noise suppression

### LINKED ISSUE #84: Feature request: Optional Noise Suppression (e.g. RNNoise)
I would like to suggest adding optional enhanced noise suppression for voice channels, possibly using something like RNNoise (open-source and lightweight).
While not as advanced as Discord's Krisp, I think it could be an improvement over the default browser/WebRTC noise suppression currently in use.

dchester:
## Summary

This adds improved noise suppression in voice channels via DTLN / machine learning

Closes #84 #263

## Additional Context

This uses the implementation from [dtln-rs](https://github.com/DataDog/dtln-rs), which anecdotally is much better than RNNoise, and very much better than the standard built-in noise suppression provided by browsers' WebRTC stacks.

DataDog wrote about the library in a blog post: https://www.datadoghq.com/blog/engineering/noise-suppression-library/

In terms of UI changes, we replace the "Noise suppression" switch with a select, where the user can choose between "None", "Standard", and "Advanced":

noise

## Considerations

The audio worklet is 8MB of wasm, taken directly from https://github.com/DataDog/dtln-rs-demo/blob/main/src/audio-worklet/dtln.js, with resampling functionality tacked onto the end.  The library's model is trained on 16kHz samples, so we downsample on the way in if need be, for browsers like Firefox who only deal in 48kHz for `getUserMedia` audio streams.

I've tested in Chrome and Firefox, but don't have easy access to test in Safari.  So that might work just fine, or might be an adventure.  I bet it will be fine, but it is untested as of yet.

Thanks for getting this project started!  Happy to take any notes you have, or do anything else necessary to get this PR into shape.


diogomartino:
8MB for a wasm is a bit insane. RNNoise wasm is 130kb for reference. How does it compare to RNNoise? Also, does it load the whole 8MB all the time to the client or only when you turn the setting on?

dchester:
In my testing so far, this library is far superior to RNNoise.

RNNoise is neat, but in practice I've found it's actually worse than nothing, because it ends up clipping actual voice audio.  So at idle, it's nice that you don't hear other people's ambient noise, but then when they start talking, it's actually harder to understand what they're saying.

This DTLN library seems to be very good at not stepping on people's actual voice audio.  You can try out a demo here: https://datadoghq.dev/dtln-rs-demo/

The 8MB only loads if/when you turn the setting on.  And then it can be cached forevermore on the device.  I think it's a few MB worth of TFLite + XNNPACK and then a few MB worth of model weights.  On the plus side though, it's pretty dang good, and fast!

diogomartino:
Tried that demo but it's not working too well for me. The processed audio is only playing in the left channel for some reason.

moyer:
Just an another point of reference: I've used RNNoise and found it to be good. Almost at parity with Krisp (which Discord uses for noise suppression).

dchester:
> The processed audio is only playing in the left channel for some reason.

Yeah, it only operates on a single channel, and that demo plays it as stereo.  

In this PR, in the mic check, we feed the mono stream into both left and right channels.  And then for the voice channel we just say it's mono if we're going through the filter.

dchester:
> 8MB for a wasm is a bit insane

For what it's worth, the Krisp model that Discord uses is ~5.8MB: https://cdn.discordapp.com/assets/krisp_browser_models/v1.0.11_1/model_32.kw

That would be in addition to whatever engine they're using for inference.  As it is in this PR, the wasm contains both the model and the engine.

diogomartino:
I'll checkout your branch and test it in Sharkord directly to see how it behaves. But in the meantime, can you pull from dev and resolve the conflicts?

diogomartino:
@dchester Just tried this, it seems the filtering is good although it really takes a hit on the CPU, in the browser task manager I was seeing 5% usage without this vs 80% usage with this. Also it adds some delay, but that's expected, I guess.

My suggestion is to close #479 and add rnnoise in this one and add both to the dropdown so users can choose.

dchester:
I'll merge RNNoise into this branch so users can choose.  

I worry a little bit that most users won't have any idea why they would choose one over the other, so that could be a point of confusion.  But at least this way, it will be easier to get feedback on both from a larger sample of users in the near term.

diogomartino:
I thought of adding a little popover that explains what are the pros and cons of each option

dchester:
@diogomartino

This branch now has RNNoise merged in as an option.



Also, regarding DTLN cpu usage, I was able to optimize by a factor of ~2x, by rebuilding the wasm with `-O3` and using a smaller quantized version of the model.  Details in `apps/client/public/dtln/README.md`

diogomartino:
Thank you! I'll take a look when I have some time.

Rand-o:
@dchester  question, Noise Suppression + DTLN  or even RNNoise, should this be paired with the other settings such as Echo Cancellation or Automatic gain control? or since we are using this we should not be using these other settings?

cgingue11:
Your write up and some quick research got me pretty excited to try DTLN. Krisp is really good and the big complaint from folks on my server is audio quality degradation coming from Discord. Now, I don't know if this is because of forced 16khz or the model quant change (or both), but I just installed v16 and I've found that DTLN has some rough results post processing. I sound like I'm coming over a radio. RNNoise sounds really crisp and works. As this was released, I don't see anyone picking DTLN over RNNoise.

Rand-o:
> Your write up and some quick research got me pretty excited to try DTLN. Krisp is really good and the big complaint from folks on my server is audio quality degradation coming from Discord. Now, I don't know if this is because of forced 16khz or the model quant change (or both), but I just installed v16 and I've found that DTLN has some rough results post processing. I sound like I'm coming over a radio. RNNoise sounds really crisp and works. As this was released, I don't see anyone picking DTLN over RNNoise.

I have the same issues and I am using RNNoise instead of DTLN.  Feedback was DTLN sounds way worse when I am speaking. Probably needs to be dialed in better on the back end.

PR #468, opened 2026-03-03
feat(466): change own user highlight slightly

### LINKED ISSUE #466: [Feature]: @ mentiones referenced to oneself should be highlighted differently
### Problem

Mentions to anyone are highlighted in a way that makes them readable. 

Mentions to you should be highlighted so they're recognizable

### Proposed Solution

Highlight with your username should have a "notification" color. With the current color scheme, maybe a pale yellow?

schemen:
## Summary
This will highlight your own user slightly differently so it becomes more recognizable. At the moment, a pale yellow.

Closes #466 

## Additional Context

This is a small change, feel free to decline if it goes against style wishes.

grafik

### DIFF
--- a/apps/client/src/components/mention-chip/index.tsx
+++ b/apps/client/src/components/mention-chip/index.tsx
@@ -1,5 +1,6 @@
-import { useUserById } from '@/features/server/users/hooks';
+import { useOwnUserId, useUserById } from '@/features/server/users/hooks';
 import { getRenderedUsername } from '@/helpers/get-rendered-username';
+import { cn } from '@/lib/utils';
 import { memo } from 'react';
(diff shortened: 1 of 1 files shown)


diogomartino:
@schemen being different is a good idea but not yellow, it's too different from the rest of the app. Maybe a white box shadow that kinda simulates a slight neon effect.

schemen:
I do have some trouble with very hard contrasts of all black and white, and personally wouldn't mind some color. I do like the nord color scheme 🙈 

In any case, how about we remove the background highlight of other names and leave it for own reference - leave the mouse over highlight of course. This would allow staying within the color scheme and make it more distinct.

diogomartino:
Nevermind, I think it's good enough.

PR #479, opened 2026-03-04
feat: use RNNoise for noise suppression

### LINKED ISSUE #84: Feature request: Optional Noise Suppression (e.g. RNNoise)

dchester:
## Summary

This adds support for noise suppression via RNNoise

Closes #84 #263
Alternative to #460

## Additional Context

This would be an alternative to the dtln noise suppression implementation in #460 which would come with some trade-offs.  I would encourage folks to test with both branches in real-world conversations and see if we think either option is much better or worse than the other in practice.

In my own testing and experience, RNNoise has the occasional tendency to clip beginnings and endings of voice audio, so you may miss the first or last syllable if the speaker "throws away" a part of a phrase or whathaveyou.  This may or may not be a big deal, depending on the context.

As far as advantages though, RNNoise is obviously significantly smaller, and it is definitely more of a known quantity that has been deployed widely in Jitsi and other projects. So it has that going for it.

The audio worklet here is by way of `@jitsi/rnnoise-wasm`, and then we tack on some neccessary bits to feed data from our `AudioContext` in the way the wasm wants it.  This is tested in Chrome and Firefox, but I don't have easy access to Safari, so it would need to be put through its paces there.


h8d13:
The main thing you'll want to look at here is not the wasm binary size/library dep but the runtime memory it eats up in your browser when using it 🤔 

Maybe that's a more relevant conversation on top of the sound suppression

diogomartino:
I even think we can add both libraries or even more. Different setups might work better with a specific algorithm. We just need to make sure the wasms aren't being loaded along with the interface since they can be quite big. They should only be loaded when a user picks it in devices and should be hard cached with the app version as a cache key, probably. But for this to be implemented correctly the way it actually plugs into your mic should be abstracted so we can add extra libs in the future easily.

The abstraction layer would also be important for the future where plugins can also manipulate mic streams.

h8d13:
Oh choice is good ! I rather just meant tested on my fork rnnoise and it eats a bit of memory 😅
Maybe that's something to profile before releasing anything, but I'm not too sure about the memory devtools

Giggitybyte:
I also think having multiple is a good idea. Sharkord really does need better noise suppression, it might be smart to merge both RNNoise and dtln now as a stop-gap. A proper API could be added down the line for more algorithms to be added through plugins.

diogomartino:
Gonna be done in #460.

PR #514, opened 2026-03-06
feat(510): add full-screen plugin slot

### LINKED ISSUE #510: [Feature]: Add FULL_SCREEN Plugin Slot with Top-Level Sidebar Navigation
### Problem

Currently, if a plugin requires a full-view canvas in the main content area, the only way to navigate to it is by having the user click the Server Name header at the top of the left sidebar. While this UX makes sense for a "Server Overview" or administrative plugin, it is highly unintuitive for "Global" or app-centric plugins (e.g., global dashboards, productivity tools, standalone mini-games, or embedded web apps). Users do not expect clicking a server's name to launch a distinct application, which limits the types of user-facing plugins developers can reasonably build.

### Proposed Solution

Introduce a new slot, PluginSlot.FULL_SCREEN, and place its navigation at the top level of the client UI (in the main left sidebar, parallel to "Direct Messages"). This allows major, app-style plugins to have dedicated, recognizable buttons and behave like native, first-class client features rather than hiding behind the server title.

Proposed Implementation Details:

Add an activePluginId state to the client's Redux store (app slice) to manage top-level view routing.

Add a dynamic PluginSidebarButtons component to the left sidebar (below "Direct Messages") that queries usePluginComponentsBySlot(PluginSlot.FULL_SCREEN).

Modify apps/client/src/screens/server-view/content-wrapper.tsx to conditionally render the active plugin component—hiding the standard <Server /> or <DirectMessages /> components—when activePluginId is set.

Additional context
I have already built and tested this implementation locally. The UI integration matches the native Sharkord sidebar button styling (hover states, active rings, layouts) perfectly. If the maintainers agree with separating "Global/App" plugins from "Server Context" plugins, I would love to submit a PR with the completed code.

Wgreen92:
## Summary

Closes #510

## Additional Context

This PR implements the following changes as discussed in the feature request:
- Added a new `FULL_SCREEN` plugin slot to allow plugins more screen real estate.
- Integrated the plugin slot with the top-level sidebar navigation for improved accessibility.
- Resolved merge conflicts with the latest `development` branch updates to `actions.ts`, `hooks.ts`, and `slice.ts`.
- All code follows the project's TypeScript and ES6+ standards.


Wgreen92:
The only thing I did notice about this implementation is when plugin is active the side bar channels are still visible (I think that's good ). But they can be selected or switched with no affect. you must exit the plugin menu to then load back into the selected channel, because it piggybacks off the Direct Messages logic.

I think a cleaner approach would be if in plugin menu, and then you click a channel, it should jump to that channel exiting the plugin. If you agree, then I can look into this implementation.

Wgreen92:
Rebased feat-510 onto the latest upstream/development to stay current with the recent releases (including the v0.0.12 tags). Verified that the full-screen plugin slot changes are intact.

apps/client/src/components/left-sidebar/plugin-buttons.tsx
```diff
@@ -0,0 +1,50 @@
...
+                const nextId = isActive ? undefined : pluginId;
+                setActivePluginId(nextId);
+              }}
+              className={cn(
+                'flex w-full items-center gap-2 rounded px-2 py-1.5 text-sm text-muted-foreground hover:bg-accent hover:text-accent-foreground',
+                isActive &&
+                  'bg-accent text-accent-foreground ring-1 ring-primary/30'
+              )}
+            >
+              <PuzzleIcon className="h-4 w-4" />
```
Giggitybyte: It seems that plugin authors can define a URI for the slot icon, but it isn't used; a puzzle icon is displayed instead.

diogomartino:
I see what you were trying to do here to keep the design of the sidebar consistent, but I'm not sure if it's better to give plugin creators the freedom to use a completely custom button. I can see cases where the plugin might need to add extra stuff to the button, and if we do it like this, they can't.

Wgreen92:
Great catch on the icon! I looked into wiring up the plugin author's custom icon, but ran into a strict architectural boundary. Currently, the join payload only sends an array of raw strings (pluginIdsWithComponents) to standard users. The actual plugin manifests (containing the logos/icons) are protected behind the plugins.get tRPC route, which throws a 403 Unauthorized for non-admin users.

To fix this securely without breaking permissions, we'd need to modify the backend's PluginManager and join route to expose a sanitized publicPluginMeta object (like sidebar: { title, icon }) to all connecting users. Since that touches the core state payload, should we tackle that backend expansion in a separate PR and keep this one focused on the FULL_SCREEN layout addition? Or if you want me to tackle it now I would be more then happy to.

We can also fallback to the current icon if one is not provided.

Giggitybyte:
>  I can see cases where the plugin might need to add extra stuff to the button, and if we do it like this, they can't.

Based off of the screenshot in the linked issue, having the button in the sidebar seems more "integrated" and app like. It feels like a part of the server.

As somebody who is wanting to implement this, I don't really see a need to add any functionality to the button itself when clicking the sidebar button would open a full screen page that's controlled by the plugin. Any sort of settings or quick actions I might want for a user could be displayed there.

If anything, perhaps a context menu for the button? Plugin authors that implement this slot could define quick links or commands which would be displayed to the user when they right click on the button.

diogomartino:
@Giggitybyte I also like t hat it feels native, but for example, in his specific case, hes building something like a media manager and he might want to add a notification badge there to show the amount of new episodes or something.

Giggitybyte:
I feel that still should be something handled inside the plugin page itself.

One of my use cases is a similar-style media manager, and I wouldn't make use of a notification badge. Users can have many libraries with varying types of content, an aggregate number next to the button wouldn't communicate much (e.g. 50 of what was added? Movies? Songs? Albums? TV Series? Anime?) 

Anything like a notification or quick information would be better displayed in the plugin page itself.  Android/iOS app launchers don't provide such customization to app developers. Discord, a more apt comparison, only allows developers of activities (a close analog to a full screen plugin) the option to specify a title and an icon. It's generally implied that an embedded app like a full screen plugin is a separate application integrated into the host app (Sharkord) and any information about the app, like notifications, can be found by opening the app itself.

Adding more information and styling to the button would just be clutter in my opinion. The type of customization/information display you're describing sounds more like a widget, instead of a button to open full screen plugin page.

Wgreen92:
@diogomartino @Giggitybyte I agree with both of you. future support for notification badge similar to # of unread dm's currently would be really nice to have for some types of plugins. I don't see myself using it on media mangers for the reason Giggitybyte mentioned. To much conflicting information for a simple number to help a user I think. Front page of my app is recently added, that's where they can see "notifications" of sorts, when they are interested.

My thoughts with current implementation are it looks native to Sharkord, which I like, it currently replicates all other tabs highlight's on hover and such. Downsides of using plugin defined icon is yeah distracting colors and such in the sidebar and breaking the cohesive feel. Also in this implementation all plugins using sidebar would have the same icon, just with different names attached. We are currently using Plugin ID for that name.

I have it currently set to SVG string I think its a perfect middle ground. An inline SVG inherits the text color from Tailwind CSS. This means if a plugin author provides an SVG, it will automatically stay monochrome, match the theme, and light up correctly on hover, preventing them from uploading a distracting, full-color PNG. I haven't looked but I wouldn't imagine it would break notifications either. Still need to implement actually using this based on what we decide.

diogomartino:
While I'm against clutter in the core of Sharkord, I'm okay for plugins to have more freedom. But we can go this route for now and in the future change the approach if needed.

apps/client/src/components/left-sidebar/plugin-buttons.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { PuzzleIcon, X } from 'lucide-react';
+import { memo, useMemo } from 'react';
+
+export const PluginSidebarButtons = memo(() => {
+  const componentsMap = usePluginComponentsBySlot(PluginSlot.FULL_SCREEN);
+  const activePluginId = useActivePluginId();
+
+  const sidebarPluginIds = useMemo(() => {
+    return Object.keys(componentsMap);
+  }, [componentsMap]);
```
diogomartino: Move this into a selector called something like `fullscreenPluginids" that uses createSelector so it's cached.

apps/client/src/features/app/slice.ts
```diff
@@ -18,6 +18,7 @@ export interface TAppState {
   browserNotifications: boolean;
   browserNotificationsForMentions: boolean;
   browserNotificationsForDms: boolean;
+  activePluginId: string | undefined;
```
diogomartino: `activePluginId` is a misleading name. Use something more specific like `activeFullscreenPluginId` so we know that what's active is the fullscreen and not the plugin itself.

packages/shared/src/plugins.ts
```diff
@@ -14,7 +14,13 @@ export const zPluginPackageJson = z.object({
...
     homepage: z.url().optional(),
     description: z.string().min(1, 'Plugin description is required'),
-    logo: z.string().optional()
+    logo: z.string().optional(),
+    sidebar: z
+      .object({
+        icon: z.string().min(1, 'Sidebar icon (SVG string) is required'),
+        title: z.string().min(1, 'Sidebar title is required')
+      })
+      .optional()
```
diogomartino: Lets keep things simple for now, make the icon the same for all plugins for now.

Wgreen92:
I’ve addressed all the feedback!

- Renamed activePluginId to activeFullscreenPluginId throughout the app for better clarity.
- Moved the fullscreen plugin ID logic into a cached selector (fullscreenPluginIdsSelector) as requested.
- Reverted the shared schema to keep the sidebar implementation simple (using the default icon/title for now).
- Resolved merge conflicts with development and fixed formatting to pass CI.

Ready for final review!

apps/client/src/components/left-sidebar/plugin-buttons.tsx
```diff
@@ -0,0 +1,48 @@
...
+                  'bg-accent text-accent-foreground ring-1 ring-primary/30'
+              )}
+            >
+              <PuzzleIcon className="h-4 w-4" />
+              <span className="flex-1 text-left truncate">{pluginId}</span>
+              {isActive && <X className="h-4 w-4" />}
+            </button>
+          </Tooltip>
+        );
+      })}
```
diogomartino: Wouldn't be bad to turn this into a separate component to keep things clean.

apps/client/src/components/left-sidebar/plugin-buttons.tsx
```diff
@@ -0,0 +1,48 @@
+import { setActiveFullscreenPluginId } from '@/features/app/actions';
+import { useActiveFullscreenPluginId } from '@/features/app/hooks';
+import { useFullscreenPluginIds } from '@/features/server/plugins/hooks';
+import { cn, Tooltip } from '@sharkord/ui';
+import { PuzzleIcon, X } from 'lucide-react';
+import { memo } from 'react';
+
+export const PluginSidebarButtons = memo(() => {
+  // Use the new cached selector hook instead of useMemo and the components map
```
diogomartino: Remove useless ai comments.

apps/client/src/components/plugin-slot-renderer/index.tsx
```diff
@@ -19,15 +19,20 @@ const SlotContextProvider = memo(({ children }: TSlotContextProviderProps) => {
 type TPluginSlotRendererProps = {
   slotId: PluginSlot;
   debug?: boolean;
+  activePluginId?: string;
```
diogomartino: Again, this name is misleading, should be changed as you did in the other one.

diogomartino (changes requested):
Also, idk why you changed the versions in the package jsons.

Wgreen92:
The version bump to 0.0.12 came through automatically during my rebase against upstream/development. It looks like it was part of the latest bot-driven chore commit.

```
Author: github-actions[bot] <github-actions[bot]@users.noreply.github.com>
Date:   Fri Mar 6 23:40:11 2026 +0000

    chore: bump version to 0.0.12

diff --git a/package.json b/package.json
index e356a55..7f0f96e 100644
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "sharkord",
-  "version": "0.0.11",
```

Working on the additional changes now.

diogomartino:
Maybe you rebased to main instead of development? When I saw your changes I noticed I had not pulled from main after the release so development was on 0.0.11 still

Wgreen92:
I've just pushed a fresh update.

- Refactored plugin-buttons.tsx: Removed the comments and extracted the button into its own internal component for a cleaner map loop.

- Naming Consistency: Updated the prop name to activeFullscreenPluginId across the renderer and wrapper.

- Cleanup: Removed the accidental local text files/workspace settings from the history.

- Rebase & Versioning: I performed a fresh rebase against the latest upstream/development. I verified that the upstream is currently at 0.0.12, so I’ve kept the version synced with that to ensure a clean merge. Ran magic.


Everything should be clean and ready for another look!

diogomartino:
I'll check this today. I'm currently making a lot of changes to the plugins so I'll merge this first so you won't have to deal with tons of conflicts.

Also the implementation might change a bit.

diogomartino:
@Wgreen92 I haven't merged this yet, but I will.

Issue #123, opened 2026-02-12
[Feature]: A way to toggle peoples webcams and screen shares.


kovila77:
As a user with bad internet connection I need a way to disable video stream.

As a temporary measure I suggest to add a button to disable all video streams.

CuteHorse:
Would be nice to have that.

AzzySama:
Yes please, we need this
Give an option to view and exit each stream

diogomartino:
This won't be implemented until `VoiceProvider` is refactored on #642.

PR #515, opened 2026-03-06
feat(123): Added button to toggle video streams

StephenArg:
## Summary
Adds a control to turn remote video streams (webcams, screen shares, external video) on and off. When disabled, the client stops consuming video producers to reduce bandwidth and CPU usage. When re-enabled, it re-consumes existing video producers.

##Changes
- Toggle button – New button in the voice control bar (left sidebar) to turn video streams on or off.
- Bandwidth savings – When video streams are off, consumers for VIDEO, SCREEN, SCREEN_AUDIO, and EXTERNAL_VIDEO are closed and tracks are stopped.
- Re-enable – Turning video streams back on re-consumes all current video producers in the channel.

##Flow
- Initial join – consumeExistingProducers uses videoStreamsEnabled to decide whether to consume video producers.
- New producer – consumeWithVideoStreamsCheck skips consuming video kinds when videoStreamsEnabled is false.
- Toggle off – stopVideoStreamConsumers() closes all video consumers; useVoiceRefs hides video in the UI.
- Toggle on – reconsumeVideoProducers() re-consumes all current video producers.

Closes #
#123

### DIFF
--- a/apps/client/src/components/top-bar/index.tsx
+++ b/apps/client/src/components/top-bar/index.tsx
@@ -9,8 +9,8 @@ import { Button, Tooltip } from '@sharkord/ui';
 import { MessageSquare, PanelRight, PanelRightClose } from 'lucide-react';
 import { memo } from 'react';
 import { PluginSlotRenderer } from '../plugin-slot-renderer';
-import { VideoController } from './video-controller';
 import { ServerSearch } from './server-search';
+import { VideoController } from './video-controller';
 import { VoiceOptionsController } from './voice-options-controller';
 import { VolumeController } from './volume-controller';
 


StephenArg:
Note the icon I used. I had a hard time deciding which to use. It may be too close to ScreenShare's icon.
Also it is neutral colored until selected, then it becomes warning red.
Clicking it again will turn it off (and turn the video streams of the remote users back on). You can perform the same effect by clicking to enable webcam or screenshare again.

CuteHorse:
Nice, but it’d be much cooler if there were a button to disable a specific stream instead of all of them.

diogomartino:
Yea, I also think you should be able to toggle specific streams as well. Also, when you are making changes to the interface, it's good to add screenshots so we have a better idea on how it looks

StephenArg:
Ok, I'll see what I can do

StephenArg:
Screenshot 2026-03-09 at 8 38 15 PM
<br>
<br>
Screenshot 2026-03-09 at 8 38 04 PM
<br>
<br>
Screenshot 2026-03-09 at 8 38 47 PM
<br>
<br>
Screenshot 2026-03-09 at 8 38 29 PM

apps/client/src/components/channel-view/voice/hooks/use-voice-refs.ts
```diff
@@ -144,10 +159,10 @@ const useVoiceRefs = (
   ]);
 
   useEffect(() => {
-    if (!screenShareStream || !screenShareRef.current) return;
```
diogomartino, line 147: Why was this first check removed?

apps/client/src/components/channel-view/voice/hooks/use-voice-refs.ts
```diff
@@ -144,10 +159,10 @@ const useVoiceRefs = (
...
+    if (!screenShareRef.current) return;
 
-    if (screenShareRef.current.srcObject !== screenShareStream) {
-      screenShareRef.current.srcObject = screenShareStream;
+    if (screenShareRef.current.srcObject !== (screenShareStream ?? null)) {
+      screenShareRef.current.srcObject = screenShareStream ?? null;
```
diogomartino, line 165: Also this doesn't seem to make much sense

apps/client/src/components/top-bar/video-controller.tsx
```diff
@@ -0,0 +1,275 @@
...
+    type,
+    isConnected,
+    onDisconnect,
+    onReconnect,
+    disabled = false
+  }: VideoStreamControlProps) => {
+    const user = useUserById(type !== 'external' ? remoteId : 0);
+
+    const Icon =
+      type === 'webcam' ? Video : type === 'screen' ? Monitor : Router;
```
diogomartino, line 49: Avoid nested ternaries, make a function like `getIcon` and call it directly or something.

apps/client/src/components/top-bar/video-controller.tsx
```diff
@@ -0,0 +1,275 @@
...
+  Button,
+  Popover,
+  PopoverContent,
+  PopoverTrigger,
+  Tooltip
+} from '@sharkord/ui';
+import { Monitor, Router, ScreenShareOff, Video, VideoOff } from 'lucide-react';
+import { memo, useCallback, useMemo } from 'react';
+
+type VideoStreamControlProps = {
```
diogomartino, line 20: Keep consistency with types, always add a T before a type, so `TVideoStreamControlProps`

apps/client/src/components/top-bar/video-controller.tsx
```diff
@@ -0,0 +1,275 @@
...
+  kind: StreamKind.VIDEO | StreamKind.SCREEN | StreamKind.EXTERNAL_VIDEO;
+  name: string;
+  type: 'webcam' | 'screen' | 'external';
+  isConnected: boolean;
+  onDisconnect: () => void;
+  onReconnect: () => void;
+  disabled?: boolean;
+};
+
+type VideoControllerProps = {
```
diogomartino, line 31: Same as above

apps/client/src/components/voice-provider/hooks/use-transports.ts
```diff
@@ -12,6 +12,16 @@ import {
 } from 'mediasoup-client/types';
 import { useCallback, useRef } from 'react';
 
+const VIDEO_STREAM_KINDS: StreamKind[] = [
+  StreamKind.VIDEO,
+  StreamKind.SCREEN,
+  StreamKind.SCREEN_AUDIO,
```
diogomartino, line 18: The variable is called VIDEO_STREAM_KINDS but includes an audio stream, is this correct?

apps/client/src/components/top-bar/video-controller.tsx
```diff
@@ -0,0 +1,275 @@
...
+          )}
+        </div>
+      </div>
+    );
+  }
+);
+
+VideoStreamControl.displayName = 'VideoStreamControl';
+
+type VideoStream = {
```
diogomartino, line 99: Same as above

apps/client/src/components/top-bar/video-controller.tsx
```diff
@@ -0,0 +1,275 @@
...
+};
+
+type VideoControllerProps = {
+  channelId: number;
+};
+
+const VideoStreamControl = memo(
+  ({
+    remoteId,
+    kind: _kind,
```
diogomartino, line 38: Why kind and type are needed here? Can't you just use kind?

apps/client/src/components/top-bar/video-controller.tsx
```diff
@@ -0,0 +1,275 @@
...
+              className={cn(
+                'flex-1 transition-all duration-200',
+                videoStreamsEnabled
+                  ? 'bg-secondary hover:bg-secondary/80 text-muted-foreground hover:text-foreground'
+                  : 'bg-red-500/15 hover:bg-red-500/25 text-red-400 hover:text-red-300'
+              )}
+              onClick={toggleVideoStreams}
+            >
+              <ScreenShareOff className="h-4 w-4 mr-2" />
+              {videoStreamsEnabled
```
diogomartino, line 237: This button shouldn't turn off your own streams.

diogomartino:
I haven't reviewed the whole code, but this pr touches the most critical and sensitive part of the app. I don't really like the way you are connecting/disconnecting consumers. I get it that it might be the best approach because it will save bandwidth but tbh in this moment adding extra complexity to voice provider might not be a good idea. That context has grown too much to the point where it's getting unmaintainable and it needs a huge refactor. I wanna avoid making changes to it unless absolutely necessary and this is not a critical feature, just a nice to have IMO. So either you change the approach and make a really simple implementation where you literally just hide the streams or we will need to rethink this in the future.

StephenArg:
> I haven't reviewed the whole code, but this pr touches the most critical and sensitive part of the app. I don't really like the way you are connecting/disconnecting consumers. I get it that it might be the best approach because it will save bandwidth but tbh in this moment adding extra complexity to voice provider might not be a good idea. That context has grown too much to the point where it's getting unmaintainable and it needs a huge refactor. I wanna avoid making changes to it unless absolutely necessary and this is not a critical feature, just a nice to have IMO. So either you change the approach and make a really simple implementation where you literally just hide the streams or we will need to rethink this in the future.

@diogomartino It is a bit much, but this was the way to do it in order to implement the feature. The people that were asking for the feature seemed to want it mainly for bandwidth reasons, so just turning off the visual while still receiving the stream defeats the point for them.

Let me know how you want to proceed. I agree personally the feature doesn’t matter much to me but this is what they asked for.

diogomartino:
I don't think the bandwidth is the most important here tbh, just the fact that the actual stream is not rendering can lower the cpu usage so that's a plus. But I agree that closing the consumer is the best approach performance wise but I don't think we should do it now.

CuteHorse:
> I don't think the bandwidth is the most important here tbh

It is very important on non broadband connections, with 4-10mbps there is not much you can do when there are two streams running, also traffic on mobile connections is almost always limited, so actually closing consumers would save a lot, and i think is the main point, not the cpu/gpu power.

diogomartino:
@CuteHorse I understand. But my original point about the implementation still stands, tho. This is more of a technical decision than a functional one.

StephenArg:
Ok, so should I close this? I'm not going to bother re-implementing it if it isn't even a feature people want.

diogomartino:
I think it's a feature people want but it's not the best time to do it...

diogomartino:
I'm gonna close this for now. This will definitely be implemented in the future.

PR #526, opened 2026-03-07
fix(501): normalize line breaks in chat messages

### LINKED ISSUE #501: [Bug]: Multi-line posts/messages periodically do not respect line breaks.
### Sharkord Version

< v0.0.11 - I believe it affects all versions to date

### Environment

Browser: Chromium based and Firefox confirmed. Possibly all browsers are affected.

As explained... At times, Sharkord, after submitting a multi-line post/message, will not respect line breaks - usually requiring a double line break in order for one to appear after submitting. 

I've had this issue appear in particular areas of a post/message, or affecting the whole post entirely - and requiring different numbers of line breaks in order to get one to appear.

Evidence 1 demonstrates how it looks after and before. Evidence 2 shows the submitted post/message, and issue from inspector's POV. Is there something up with the 'hard-break' class for line breaks?

Image

Image

### Steps to Reproduce

It's not always reproducable. It appears to affect a post/message at random. At least I have not been able to consistently trigger it.

### Expected Behavior

A line break to be visible in the submitted post where a break was entered in the post/message text editor.

### Logs / Screenshots

_No response_

dchester:
## Summary

With this change, we normalize empty lines on the way out of the editor, so that they will round-trip and preserve whitespace in the expected way.

Closes #501

## Additional Context

This works around an issue with prosemirror, where they inject `<br class="Prosemirror-trailingBreak" />` at the end of block elements, only while editing, only so you can get your cursor into that place.  That element they inject is not meant to actually add a visual rendered line, but alas, it does in some cases, and so that is what creates the disconnect.  When we render the message outside the editor, nobody is injecting that `br` and so empty lines that were there in the editor appear to be missing now.

Our "least-bad" workaround approach here is to find block elements with trailing `<br class="hard-break">`, with that block element followed by another block element -- and then inject a `<p></p>` between the block elements and delete the `<br>`.  Essentially, we "promote" the empty line at the end of the block to be its own `p` sibling to the previous and next elements.  Additionally, we also make sure to translate "disallowed" block elements into `p` elements, to handle case 1 below.

Also, the `inline` treatment meant for the `(edited)` indication was getting in the way, because for non-edited messages that `nth-last-child` selector was still getting applied and making some block elements to be rendered as inline!

## Reproduction Steps

There are at least two ways to reproduce this bad behavior, both requiring editing copy/pasted rich content in the editor.

1. Copy the rendered "line 1 / line 2" text below
2. In the chat input, place your cursor at the end of the first list item and press `shift`+`enter`
3. Observe there is a blank line between 1 and 2 and press `enter` to send

The rendered message now contains no blank line between 1 and 2.

#### Copy/paste these two lines for each case below in order to reproduce the bug...

Case 1 - (`h4` followed by `p`):
> #### line 1
> line 2

Case 2 -  (`p` followed by `p`)

> line 1
>
> line 2

## Potential Alternatives

This PR seems like it will be net-positive as-is, but it may be worth considering some alternatives.

**Markdown plain text input** - Using a plain `textarea` (or `contenteditable="plaintext-only"`) might just make everyone's life simpler and better.  Web-based WYSIWYG editors are so extremely difficult that for example, even after ten years of smart people trying with all their might, ProseMirror still has bugs like this.  Even when they work as intended, humans have a hard time with them rich-text editors anyway.  In the simpler version, messages could be stored in plain text and interpreted as markdown on the way to be rendered.  This would remove a ton of complexity and computational overhead while editing on low-powered devices.

**No hard breaks** - Right now, pressing `shift`+`enter` is the way to make a newline without sending the message -- but the editor is *also* interpreting that `shift` to mean "make a hard break within this block" rather than "make a new block".  So we might try overriding the keymaps to say that from the editor's point of view, `shift`+`enter` should be treated as `enter`.  This is probably a good idea on its own, but it's a bigger change, and still doesn't fix the variation of this collapsing newline bug where a hard break is part of the rich content that is pasted in and then edited.

## See also

Some discussion in upstream projects:
- https://github.com/ProseMirror/prosemirror/issues/1511
- https://github.com/ueberdosis/tiptap/issues/1500

## Summary by CodeRabbit

* **New Features**
  * Added visual indicator to show when a message has been edited.

* **Bug Fixes**
  * Fixed message formatting to properly preserve line breaks and empty paragraphs.
  * Improved HTML processing to maintain headers and list items as content blocks instead of losing them.

### DIFF
--- a/apps/client/src/components/channel-view/text/index.tsx
+++ b/apps/client/src/components/channel-view/text/index.tsx
@@ -14,7 +14,7 @@ import {
   ChannelPermission,
   TYPING_MS,
   getTrpcError,
-  linkifyHtml
+  prepareMessageHtml
 } from '@sharkord/shared';
 import { Spinner } from '@sharkord/ui';
 import { throttle } from 'lodash-es';
@@ -93,7 +93,7 @@ const TextChannel = memo(({ channelId }: TChannelProps) => {
 
       try {
         await trpc.messages.send.mutate({
-          content: linkifyHtml(message),
+          content: prepareMessageHtml(message),
           channelId,
           files: files.map((f) => f.id)
         });
--- a/apps/client/src/components/channel-view/text/message-edit-inline.tsx
+++ b/apps/client/src/components/channel-view/text/message-edit-inline.tsx
@@ -1,6 +1,10 @@
 import { TiptapInput } from '@/components/tiptap-input';
 import { getTRPCClient } from '@/lib/trpc';
-import { type TMessage, isEmptyMessage, linkifyHtml } from '@sharkord/shared';
+import {
+  type TMessage,
+  isEmptyMessage,
+  prepareMessageHtml
+} from '@sharkord/shared';
 import { AutoFocus } from '@sharkord/ui';
 import { memo, useCallback, useState } from 'react';
 import { toast } from 'sonner';
@@ -29,7 +33,7 @@ const MessageEditInline = memo(
         try {
           await trpc.messages.edit.mutate({
             messageId: message.id,
-            content: linkifyHtml(newValue)
+            content: prepareMessageHtml(newValue)
           });
 
           toast.success('Message edited');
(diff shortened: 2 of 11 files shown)


diogomartino:
Gave it a quick look and I think it looks good.

@coderabbitai review this pr.

Issue #472, opened 2026-03-03
[Feature]: Be able to close a PM


6004cc0-cmyk:
### Problem

As of now, you can't close a PM with another user.

### Proposed Solution

Add a close link to the right of the user in the left panel's list. When clicking on it, close the PM and remove the user from the PM list.

Optional, but recommended for best privacy: when clicking on close, show a modal asking the user if he or she wants to delete the content of the PM and only for self. Add a checkbox stating "Delete the chat data for [username here], too". If the checkbox is checked, delete the content for the other user too.

ShiningVenus:
My idea was to when your cursor hovers over the DM itself an X would appear, you click it, and it is closed. The rest you added does sound quite nice too to have. So for example if I hovered over your username in a PM/DM an X would appear. I'd click it and it'd close the DM/PM we could either allow a pop up ensuring a user is sure or not and opt to never show again, perhaps.  The X would be on the right side and maybe just a easy to view/see color or icon? I didn't wanna open a new issue since this one was made. I hope this makes some sense?

6004cc0-cmyk:
I would rather say that the X would always be visible (where you mentioned it). This way the X will always be visible for all devices, even portable ones like smartphones.

ShiningVenus:
> I would rather say that the X would always be visible (where you mentioned it). This way the X will always be visible for all devices, even portable ones like smartphones.

Yes I agree that the X should always be visible and there regardless but maybe again a one time prompt which you can opt to turn off forever when it asks "are you sure you meant to close this DM?" or a similar message. Also maybe in one prompt or so it can ask about your optional ideas like deleting the data of the DM/PM etc.

diogomartino:
The deletion is not good since the other user might not want to lose the data. Closing is fine and just means it won't show up in your conversations until you open it again.

6004cc0-cmyk:
Sure. I understand that, but here's another solution. If 1 user closes a DM, leave the chat history to the other user. But when that user also closes the DM, delete the chat history all together. With other words, if both parties closes a DM, delete the data for that DM.

diogomartino:
I think that's adding complexity where it's not needed. Is that a common thing in chat apps?

6004cc0-cmyk:
It should be a common thing in all chat apps. If not, why are the chat history that no one needs still in the database?

Another reason of the solution is to keep everything clean. Do not keep data that are not used by anyone anymore.

Kanda:
[...]

PR #532, opened 2026-03-08
feat: show channel name as placeholder text in compose input

### LINKED ISSUE #531: [Feature]: Add placeholder text for chat message input
### Problem

When going to compose a chat message, the user could benefit from an affirmation/reminder about who is the audience who will receive their message.

### Proposed Solution

Other platforms like Slack and Discord put placeholder text in the chat message input, like "Message #General" for example, if they are about to send a message to the #General channel.

> Discord's input with placeholder text

dchester:
## Summary

This change adds descriptive placeholder text in the chat message composition input. 

Closes #531 

## Additional Context

We follow suit with Discord, Slack, and others, adding "Message #General" for example, if we are about to send a message to the #General channel.  For voice channels, we omit the `#` prefix, like Discord does.  For DMs, we say like "Message @Arthur", if we are about to send a message to Arthur.

In order to have the name of the DM recipient, we move the conversations list to the Redux store, rather than just fetching it locally in the sidebar component.  This should also help if/when we want to show that user's name in the top bar header of the conversation, that currently just says, "Direct Message"


Giggitybyte:
I feel this should be worded differently as text channels in Sharkord can contain spaces and uppercase letters. `Message #long-channel-name` might look okay in Discord, but based off your screenshot, `Message #Long Channel Name` doesn't seem right.

Something like `Send message to {channel_name}` or `Send a message in {channel_name}` would read better to me.

Example from a Discord thread:
Screenshot_20260308_140041

diogomartino:
I agree with @Giggitybyte, that looks better. Also, 90% of your changes have nothing to do with the actual issue so don't know why you made all those changes.

dchester:
@diogomartino

Yes, I'll change to `Send a message in {channel_name}`.

> 90% of your changes have nothing to do with the actual issue

Did you see my note above?  All of these changes are in service of this feature.

> In order to have the name of the DM recipient, we move the conversations list to the Redux store, rather than just fetching it locally in the sidebar component. This should also help if/when we want to show that user's name in the top bar header of the conversation, that currently just says, "Direct Message"

Do we already have access to the name of the DM recipient in some way that I missed?  Or would you rather achieve this in a different way?  For a second I thought about parsing user ids out of the channel name, but that seemed fragile.

diogomartino:
Oh sorry, didn't see that. But still, if you want to move things to the store, do it all the way and keep the conversations there for good (as messages, channels, etc.). Pushing an event so that you can refetch the conversations doesn't make sense.

dchester:
@diogomartino 

Okay, this is saner now.  We fetch the full list of dm conversations just once in the initial connection, and then keep in sync from there with adds/updates for brand new conversation threads and updated times.

There are a few new hard coded user facing strings in here that don't have i18n entries.  Do you want to fill that in as part of this PR, or can that be follow-up?

diogomartino:
I'm gonna check this when I have some time, but if you could do the i18n change that would be great.

diogomartino:
The DM's implementation is not the best IMO, there are certain details you missed. I'll probably need to change some things in DM's when I do this: https://github.com/Sharkord/sharkord/issues/472 and then this will become much easier. Until then, we have two options:
- close this pr
- remove all DM's logic present here and just apply the changes in normal text channels

dchester:
This now has no special handling for DMs and just does the contextual placeholder for channels.

diogomartino:
Thank you, I will check this again when I have some time.

diogomartino (approved):
LGTM

PR #535, opened 2026-03-08
feat(533): make chat input max height configurable by draggable divider

### LINKED ISSUE #533: [Feature]: Let the chat input grow as needed by default
### Problem

When typing a chat message longer than a few lines, it is awkward to have to click the chevron to expand it upwards.

### Proposed Solution

Let the input grow in height as needed (up to the max height), as it does in Discord and other similar apps.  Maybe the expand/collapse toggle was serving a purpose at some point in time?  But as of now, it seems we lose nothing if we remove the button and always go with the "expanded" behavior.

dchester:
## Summary

In this change we remove the expand/collapse toggle on the chat input, and let the height grow as needed right from the start.

Closes #533

### DIFF
--- a/apps/client/src/helpers/storage.ts
+++ b/apps/client/src/helpers/storage.ts
@@ -27,6 +27,8 @@ export enum LocalStorageKey {
   BROWSER_NOTIFICATIONS = 'sharkord-browser-notifications',
   BROWSER_NOTIFICATIONS_FOR_MENTIONS = 'sharkord-browser-notifications-for-mentions',
   BROWSER_NOTIFICATIONS_FOR_DMS = 'sharkord-browser-notifications-for-dms',
+  CHAT_INPUT_HEIGHT_VH = 'sharkord-chat-input-height-vh',
+  THREAD_INPUT_HEIGHT_VH = 'sharkord-thread-input-height-vh',
   BROWSER_NOTIFICATIONS_FOR_REPLIES = 'sharkord-browser-notifications-for-replies',
   LANGUAGE = 'sharkord-language',
   PLUGIN_SLOT_DEBUG = 'sharkord-plugin-slot-debug',
--- a/apps/client/src/components/channel-view/text/chat-input-divider.tsx
+++ b/apps/client/src/components/channel-view/text/chat-input-divider.tsx
@@ -1,0 +1,172 @@
+import {
+  type LocalStorageKey,
+  removeLocalStorageItem,
+  setLocalStorageItem
+} from '@/helpers/storage';
+import { useCallback } from 'react';
+
+const MAX_VH = 80;
+const MIN_PX = 56;
+const RESET_THRESHOLD_PX = 10;
+
+type TChatInputDividerProps = {
+  composeContainerRef: React.RefObject<HTMLDivElement | null>;
+  scrollToBottom: () => void;
+  isAtBottom: () => boolean;
+  storageKey: LocalStorageKey;
+  defaultMaxHeightVh: number;
+};
+
+// calculate the minimum acceptable chat input height
+export const measureMinHeight = (composeEl: HTMLDivElement): number => {
+  const proseMirror = composeEl.querySelector(
+    '.ProseMirror'
+  ) as HTMLElement | null;
+  if (!proseMirror) return MIN_PX;
+
+  // clamp to one line to measure empty state
+  const savedHeight = composeEl.style.height;
+  const savedMaxHeight = composeEl.style.maxHeight;
+  const scrollRow = composeEl.querySelector(
+    '.compose-scroll-row'
+  ) as HTMLElement | null;
+  const savedScrollTop = scrollRow?.scrollTop ?? 0;
+  proseMirror.classList.add('line-clamp-1');
+  composeEl.style.height = '';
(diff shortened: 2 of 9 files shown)


dchester:
This branch now has a setting for the max height of the chat input.

The max height as a percentage is inspired by Discord's behavior.  With much content in the input, if you have a tall browser window, or a short one, or if you zoom in/out, the percentage split remains constant.  Which seems to make sense, whether you have a giant 8k screen, or are constrained on a laptop or smaller device.



We also have a tweak to the `ResizeObserver` so that when the chat input grows as you type more lines, it no longer appears to "cover" the most recent messages in the chat content above.  This is also matching Discord.

@diogomartino what do you think of this?  I think it address the use cases we want to cover -- gives complete control for however the user wants to allocate their real estate, while also working in a familiar way -- but I'm not sure if you may have had something different in mind when you said we should "make it an option".

diogomartino:
Need to check this when I have some time.

diogomartino:
Revisiting this I don't think it makes sense to have a setting for that. The more user friendly approach would be to make the actual input resizable and make that value persistent.

apps/client/src/helpers/storage.ts
```diff
@@ -115,11 +138,13 @@ export {
   getLocalStorageItemAsJSON,
   getLocalStorageItemAsNumber,
   getLocalStorageItemBool,
+  getLocalStorageItemNumber,
```
diogomartino: This already exists as `getLocalStorageItemAsNumber`.

diogomartino (changes requested):

dchester:
We discussed the "resizeable" option a bit in #533, where we'd have a draggable divider between the chat messages and input:

> You could drag it up/down and it would stay there until you submit your message. Then when you submit, it would collapse back down to one line. Then when you type a long message again, it would auto-grow to wherever you dragged the height to last.

We talked about some pros and cons in that thread, but I think it could be worth a try.  Is that like what you're thinking?

apps/client/src/helpers/storage.ts
dchester: Ah, good point.

diogomartino:
Yes, exactly.

dchester:
I've pushed up an iteration with a draggable divider that lets the user size up and down the chat input.  No more slider in settings.


https://github.com/user-attachments/assets/af7cf5fb-7ffd-4b60-a697-5b538805fed6



### Behavior

- The input grows as more lines are entered into the input, up to a max height, default `35vh`

- We can drag only up to a sensible maximum (`80vh`) and down to a sensible minimum (one line of text)

- When the user drags the divider, their `dragend` height becomes the new max height; except...

- When the user drags the divider all the way to the minimum height, we treat that as a "reset" to the default

- Upon sending a message, the input returns to a height of one line, and will again grow as needed up to the max-height

### A few minor structural changes

- We move the Emoji button out from the tiptap input up into the message composer, so that it can sit next to the Paperclip and Send buttons in the DOM, and not scroll with the editor

- We move the typing indicator above the uploads so that any indications of typing don't appear interspersed between the uploads and the message being composed



> Example showing an uploaded file, a typing indicator, and a reply banner, all at once!

### Test cases

- Typing indications should be in order
- Space reserved at the end of messages for the typing indicator
- Emoji picker should be functional
- Attachments should be as they were
- Message composer should appear disabled while uploading is in progress
- Scrollbar should appear at the right edge when content exceeds max height
- Emoji, Paperclip, Send buttons should not be able to scroll away
- Messages should stay pinned to bottom when input grows, by typing or dragging

diogomartino:
One problem I found while testing is that when you have files in the message, you can resize down to the point you don't see the chat anymore. This will be more problematic with the new file previews because they are bigger.

Also I don't think you should reset the size when you send a message. If you set it to a specific height, it's because you want it like that. I also think the height should persist between channels.

One other problem, inside threads currently it's not resizable and it should be, the message compose experience should be the same in all places.

Apart from that I think it looks pretty good. Haven't checked the new code yet tho, please first address these.

dchester:
@diogomartino Your points above should be addressed now.  

You cannot resize down too far when you have files; the height you drag to sticks even after submitting your message; and it all works the same way for thread replies.

diogomartino:
I can see some conflicts, do you have the file preview changes?

dchester:
Yes, file preview changes are in there.  The conflict was just from yesterday's "tweak messages render" commit dbee5c1.  I merged that in now, and we have green checkmarks again.

diogomartino (approved):
Looks good, ty!

diogomartino:
@dchester I was making some code formatting (blame my OCD) and I found a small problem:

On smaller screens, when you have a lot of files and you try to move the divider, the input becomes scrollable and things jump around. 

https://github.com/user-attachments/assets/90bd97c6-02b8-4069-90ae-ed4f5eb0ca37

Since I don't have time atm to debug this and you should know your code, if you can check it it would be great. If not, I'll investigate myself later without a problem. Ty!

I pushed the formatting to dev and I moved a couple of functions around, but I didn't touch the logic.

dchester:
Nice, I'll track down that issue with small screens + lots of files.

PR #541, opened 2026-03-09
feat(401): support basic markdown in chat

### LINKED ISSUE #401: [Feature]: Basic Markdown

dchester:
## Summary

This adds robust support for basic markdown in chat

Closes #401
Alternative to #429 

## Additional Context

In this implementation, the input acts as a markdown-aware plain text editor.  Markdown characters have a subtle syntax-highlighting type of visual acknowledgement. Some screenshots of the composition editor and rendered messages:

composing

> Composing a message

rendered

> A rendered message with inline styles

blocks

> A message with block elements

### Markdown / HTML conversion

When a user types some input and presses send, we render the markdown to html, and the html is what gets saved and broadcast.  When a user clicks to edit their already-sent message, we convert the html back to markdown, they make their edits, and then we convert back to html.  When a user pastes rich content it's a similar flow, wherein we convert to markdown on the way into the editor.

### Test cases

- We should be able to copy/paste rich text into the editor, and end up with the pasted contents having a best-effort markdown interpretation.  Headings, links, line breaks, code, etc should all flow through

- Composing a message, sending, then editing that sent message should round-trip back into the editor as it was before it was sent.  That is, modulo a few non-deterministic cases such as how markdown allows for either `-` or `*` for list item bullet point characters.

- Other editor features like :emoji: and @mentions and /commands should be unaffected. 

- We should be able to chat `I like <p> tags` and expect that to go through as-is, without trying to put an actual paragraph tag in there.

- We should be able to paste in source code snippets into codeblocks and have the leading whitespace on each line be preserved


diogomartino:
I'm not a fan of handling markdown -> html conversions like that. We should use this: https://tiptap.dev/docs/editor/markdown.

dchester:
I'll try a version that uses the tiptap markdown extension.

dchester:
@diogomartino This now uses the tiptap markdown extension.  What do you think?

diogomartino:
Need to check this when I have more time. But from a glance I think it has too much going on for an extension that should be plug and play, but I might be wrong.

dchester:
It's true, there is a lot going on.  I pushed up an iteration that I think is a modest improvement, but I don't think we can make it drastically simpler, if we want to stick with this UX, and also stick with tiptap.  

In its heart, tiptap is a rich text WYSIWYG editor, and so it takes some coaxing to get it to behave as a plain-text syntax-aware markdown editor.  The tiptap markdown extension only really facilitates converting between md <=> tiptap object model <=> html.  It's just a low-level building block, not any sort of markdown editing experience, so there's still a fair amount of maneuvering we have to do to get what we want.

I made a diagram for how this is working...


All that said though, I think this technical approach is fine enough, and would be a very nice user experience step forward for sharkord.

But if it's really too much, we could look for ways to either cut UX scope, or else look at something like codemirror instead of tiptap, which is more directly built for this type of editing mode.  But of course all that would come with its own set of trade-offs.

Ryah:
Any update on this?

PR #554, opened 2026-03-11
feat(215): add voice user volume controls in sidebar

### LINKED ISSUE #215: Request: Right-click menu to adjust user volume
Currently the only way to adjust the volume of the other people in a call is by going to the call and then using the menu at the top.
It would be nice to be able to adjust a users volume by right-clicking on their name in the call on the left sidebar, similar to how it works in Discord.

univeous:
## Summary

Closes #215

This PR adds a right-click volume control for users in the voice channel member list (left sidebar).

- Add a right-click context menu on voice user entries with a mute toggle and volume slider
- Show a compact volume indicator next to the user’s status icons when volume is not 100%
- Add i18n label for the user volume control

### DIFF
--- a/apps/client/src/components/left-sidebar/voice-user-context-menu.tsx
+++ b/apps/client/src/components/left-sidebar/voice-user-context-menu.tsx
@@ -1,30 +1,34 @@
 import { UserAvatar } from '@/components/user-avatar';
-import { useVolumeControl } from '@/components/voice-provider/volume-control-context';
+import { useUserVolumeControl } from '@/components/voice-provider/hooks/use-user-volume-control';
 import type { TVoiceUser } from '@/features/server/types';
-import { Button, ContextMenu, ContextMenuContent, Slider } from '@sharkord/ui';
+import {
+  Button,
+  ContextMenu,
+  ContextMenuContent,
+  ContextMenuTrigger,
+  Slider
+} from '@sharkord/ui';
 import { Volume2, VolumeX } from 'lucide-react';
 import type { ReactNode } from 'react';
 import { useTranslation } from 'react-i18next';
 
 type TVoiceUserContextMenuProps = {
   user: TVoiceUser;
-  trigger: ReactNode;
+  children: ReactNode;
 };
 
 export const VoiceUserContextMenu = ({
   user,
-  trigger
+  children
 }: TVoiceUserContextMenuProps) => {
   const { t } = useTranslation('sidebar');
-  const { getUserVolumeKey, getVolume, setVolume, toggleMute } =
-    useVolumeControl();
-  const volumeKey = getUserVolumeKey(user.id);
-  const volume = getVolume(volumeKey);
-  const isMuted = volume === 0;
+  const { volume, isMuted, setVolume, toggleMute } = useUserVolumeControl(
+    user.id
+  );
 
   return (
     <ContextMenu>
-      {trigger}
+      <ContextMenuTrigger>{children}</ContextMenuTrigger>
       <ContextMenuContent className="w-64">
         <div className="px-3 py-3">
           <div className="flex items-center gap-2 mb-3">
@@ -38,7 +42,7 @@ export const VoiceUserContextMenu = ({
             <Button
               variant="ghost"
               size="sm"
-              onClick={() => toggleMute(volumeKey)}
+              onClick={toggleMute}
               className="h-6 w-6 p-0"
             >
               {isMuted ? (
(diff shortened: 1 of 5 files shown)


apps/client/src/components/left-sidebar/voice-user.tsx
diogomartino, line 1: Split this component into two. The context menu can live in a different component.

Also use `useIsOwnUser` instead of doing it manually.

univeous:
I agree the volume indicator might be a bit noisy. However, I think the mute icon is worth keeping - if a user is muted but there's no UI indication, it would be quite confusing.

apps/client/src/components/left-sidebar/voice-user-context-menu.tsx
```diff
@@ -0,0 +1,69 @@
...
+import type { TVoiceUser } from '@/features/server/types';
+import { Button, ContextMenu, ContextMenuContent, Slider } from '@sharkord/ui';
+import { Volume2, VolumeX } from 'lucide-react';
+import type { ReactNode } from 'react';
+import { useTranslation } from 'react-i18next';
+
+type TVoiceUserContextMenuProps = {
+  user: TVoiceUser;
+  trigger: ReactNode;
+};
```
diogomartino, line 18: Just a little thing, in react when you are trying to do this kind of thing, the best practice is to pass whatever you want the trigger to be as a children since it will look cleaner when you reuse the component.

apps/client/src/components/left-sidebar/voice-user-context-menu.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const { t } = useTranslation('sidebar');
+  const { getUserVolumeKey, getVolume, setVolume, toggleMute } =
+    useVolumeControl();
+  const volumeKey = getUserVolumeKey(user.id);
+  const volume = getVolume(volumeKey);
+  const isMuted = volume === 0;
+
+  return (
+    <ContextMenu>
+      {trigger}
```
diogomartino: So here you will render 

```
<ContextMenuTrigger>
   {children}
</ContextMenuTrigger
```

apps/client/src/components/left-sidebar/voice-user.tsx
```diff
@@ -1,63 +1,93 @@
...
-          className="h-5 w-5"
-          showUserPopover={true}
-          showStatusBadge={false}
-        />
+const VoiceUser = memo(({ user, isOwnChannel = false }: TVoiceUserProps) => {
+  const isOwnUser = useIsOwnUser(user.id);
+  const { getUserVolumeKey, getVolume } = useVolumeControl();
+  const volumeKey = getUserVolumeKey(user.id);
+  const volume = getVolume(volumeKey);
+  const isMuted = volume === 0;
```
diogomartino: This would be better of in a separate hook called, like:

`const { volume, isMuted } = useUserVolumeControl(user.id)`

When you do this, also check for other places in the app where you can reuse this hook

apps/client/src/i18n/locales/en/sidebar.json
```diff
@@ -67,4 +67,6 @@
   "deleteCategoryMsg": "Are you sure you want to delete this category? This WILL delete all the channels within this category. This action cannot be undone.",
   "categoryDeleted": "Category deleted",
   "failedDeleteCategory": "Failed to delete category"
+  ,
```
diogomartino: This looks odd, not sure if it's just a bug in github

diogomartino (changes requested):

diogomartino:
Yes, you can leave the muted thing just make sure users can easily get if the user was muted by you OR if the actual user muted himself.

univeous:
> Yes, you can leave the muted thing just make sure users can easily get if the user was muted by you OR if the actual user muted himself.

- User muted themselves: a muted microphone icon
- I muted the user: a muted speaker icon
I think that's enough to distinguish between the two.

sc

univeous:
Made changes based on feedback.
I'm not sure whether I should refactor `VolumeButton` to use the new hook — it doesn't seem to be exclusively for user volume, but also handles screen share volume and similar cases. I'll leave it as-is for now.

apps/client/src/components/voice-provider/hooks/use-user-volume-control.ts
```diff
@@ -0,0 +1,17 @@
+import { useVolumeControl } from '../volume-control-context';
+
+export const useUserVolumeControl = (userId: number) => {
+  const { getUserVolumeKey, getVolume, setVolume, toggleMute } =
+    useVolumeControl();
+  const volumeKey = getUserVolumeKey(userId);
+  const volume = getVolume(volumeKey);
+  const isMuted = volume === 0;
+
+  return {
```
diogomartino, line 10: This should be memoized.

diogomartino (approved):
I'm gonna merge and make some small adjustments.

PR #567, opened 2026-03-12
feat(566): add customizable date and time format settings

### LINKED ISSUE #566: [Feature]: Customizable date and time format settings
### Problem

Sharkord uses fixed timestamp formats that cannot be customized. Users cannot choose their preferred date/time format (MM/DD vs DD/MM, 12h vs 24h) or switch between relative ("2 hours ago") and absolute timestamps.

### Proposed Solution

Add user-configurable date and time format settings:

**Format Selection:**
- Preset format options with localized examples
- Custom format input with live preview
- Separate date and time format controls

**Display Preferences:**
- "Always show absolute time" toggle to disable relative timestamps
- Hover over relative time to show absolute time

univeous:
## Summary

Closes #566

This PR adds user-configurable date and time format settings:
- Added options to choose date and time format (with presets and custom format support)
- Added an option to always show absolute time instead of relative time
- Added hover on relative time to view absolute time

The hover functionality required refactoring the `RelativeTime` component from a render prop pattern to directly returning a span element, which resulted in changes across multiple files. @diogomartino what do you think about this approach?

## Additional Context

The changes include:
- New `DateFormatSettings` component in user settings
- New `useDateFormat` hook using `useSyncExternalStore` for reactive localStorage
- Refactored `RelativeTime` component to support hover and preferAbsoluteTime
- i18n translations for English and Chinese

sc

### DIFF
--- a/apps/client/src/components/channel-view/text/messages-group.tsx
+++ b/apps/client/src/components/channel-view/text/messages-group.tsx
@@ -7,7 +7,6 @@ import {
   DELETED_USER_IDENTITY_AND_NAME,
   type TJoinedMessage
 } from '@sharkord/shared';
-import { format } from 'date-fns';
 import { memo } from 'react';
 import { Message } from './message';
 
@@ -47,11 +46,8 @@ const MessagesGroup = memo(
               {getRenderedUsername(user)}
             </span>
             <RelativeTime date={date}>
-              {(relativeTime) => (
-                <span
-                  className="text-primary/60 text-xs"
-                  title={format(date, 'PPpp')}
-                >
+              {(relativeTime, hoverProps) => (
+                <span className="text-primary/60 text-xs" {...hoverProps}>
                   {relativeTime}
                 </span>
               )}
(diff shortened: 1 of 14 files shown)


univeous:
If this gets merged, the `Others` settings will have quite a bit of content. Perhaps we could consider adding an `Interface` tab; of course, that's beyond the scope of this PR.

diogomartino:
Tbh I don't think this is needed. I see you are opening a lot of pr's but I really advise you to wait until the issue is tagged with planned before you open a PR, otherwise it gets really hard for me to keep things on track.

But anyway:
- custom formats is too much, gonna create more problems than it solves
- the absolute date switch is ok
- the changes to RelativeTime are not ok, the component was made like that to allow you to use anything as children, forcing a span completely misses the point of it
- you can create a new tab called interface yes, it will look better

Do this and we'll go from there.

univeous:
Sorry for the flood of PRs — I realize it's not ideal for your planning workflow, and I'll be more mindful of that going forward. For feature PRs, I'll wait until there's a consensus before opening one.

Just to make sure I'm understanding correctly: custom format presets are fine, but the free-form "custom" format string input at the end of the options is unnecessary and should be removed? Let me know if I'm off.

diogomartino:
Revisited this and I think this is not worth adding. Sorry about any lost work here.

PR #571, opened 2026-03-12
feat(231): add speaking effect on left sidebar

### LINKED ISSUE #231: Request: Hightlight username/avatar when a user is speaking
Add a highlight outline to the username and avatar when someone speaks in the call on the sidebar. It currently does this to the cards in the call but if you are viewing a different text channel you can't see that.

diogomartino:
## Summary

Closes #231

## Additional Context

image


ivan-vrabac:
respect it pls, so nice

Rand-o:
I updated to release 15 however I dont see any of these animations, anything else I need to do?

Rand-o:
> I updated to release 15 however I dont see any of these animations, anything else I need to do?

It was a chrome extension that somehow was  messing with it.... user agent extension I had that I have now removed :)

PR #598, opened 2026-03-17
feat(595): add rename action to user management

### LINKED ISSUE #595: [Feature]: Rename option for user management
### Problem

I would like my own server to ensure that everyone follows the appropriate naming rules for usernames, and to make moderation easier in case of inappropriate or offensive expressions.

### Proposed Solution

In addition to the kick, ban, and delete options, there should also be a “rename” button in user management so that a user’s name can be changed if necessary.

gf-svg:
Closes #595

## Summary
- add Rename action to user management
- allow moderators/owners to rename a user from the management UI
- update related UI state and validation
- add a new permission to roles

image
image
image


diogomartino:
I'm gonna close this, my reasons are in the issue itself. But in summary, moderation as a whole will need a general overhaul, and adding specific things like this will make it harder in the future.

PR #617, opened 2026-03-22
feat: add vertical layout option for portrait monitors

daelsc:
Closes #613

## Summary
- Adds a "Vertical layout (portrait mode)" toggle to the Voice Options popover
- When enabled, forces the voice grid to a single column, stacking all stream cards vertically
- Ideal for widescreen monitors in portrait orientation
- Setting is persisted to localStorage

## Implementation
Follows the existing pattern used by `hideNonVideoParticipants`:
- LocalStorageKey, Redux slice state + reducer, selector, hook, action, UI toggle
- VoiceGrid accepts a `verticalLayout` prop that overrides column calculation to `grid-cols-1`
- Also applies to the pinned card thumbnail bar column count
- Includes en + zh i18n strings

## Test plan
- [x] Join a voice channel with multiple participants
- [x] Toggle "Vertical layout (portrait mode)" in Voice Options
- [x] Verify all cards stack in a single column
- [x] Verify pinned card view also respects the setting
- [x] Verify setting persists across page reloads
- [x] Verify disabling the toggle restores the default grid layout

PR #621, opened 2026-03-24
feat(613): responsive grid layout in voice channels

### LINKED ISSUE #613: [Feature]: Support for Vertical Monitor Orientation
### Problem

Currently when in a voice channel with multiple other users streaming it arranges the video strems in a way that assumes a traditional landscale monitor orientation.  This makes for very inefficient use of screen real estate due to massive letterboxing.

### Proposed Solution

Detect the screen orientation and arrange video streams vertically when in vertical mode.  Or give the user some control over how the stream windows are arranged.

dchester:
## Summary

This makes the layout of cards in video channels responsive to the dimensions of the viewport.

Closes #613
Alternative to #617 

## Additional Context

https://github.com/user-attachments/assets/33fc3a0d-c0e7-4d78-87da-37522bd20242

In this approach, we compute the number of rows and columns that will result in cards having the most natural aspect ratio.  If the space is very narrow, cards end up in a column.  If the space is very wide, cards end up in a row.

Screenshot from 2026-03-23 23-48-12

Screenshot from 2026-03-23 23-48-29

Screenshot from 2026-03-23 23-48-50

Screenshot from 2026-03-23 23-49-13

### Scrolling fix

This PR also includes a fix for a rough bug when a card is pinned, where if there are enough participants in the channel, some can flow off the left of the scrolling bar at the bottom so as to be completely unreachable.  The fix is to use `justify-center-safe` instead of `justify-center`.

Screenshot from 2026-03-23 23-55-00

> Without this fix, it is impossible to scroll the left-most item into view

Let me know if you want this to be separated out, but it's just a change of a css class name.

### DIFF
--- a/apps/client/src/components/channel-view/voice/voice-grid.tsx
+++ b/apps/client/src/components/channel-view/voice/voice-grid.tsx
@@ -1,5 +1,13 @@
 import { cn } from '@/lib/utils';
-import { isValidElement, memo, useMemo, type ReactNode } from 'react';
+import {
+  isValidElement,
+  memo,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+  type ReactNode
+} from 'react';
 
 type TVoiceGridProps = {
   children: ReactNode[];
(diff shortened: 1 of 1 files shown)


daelsc:
this looks great!  better than my solution!

diogomartino:
Looks great, gonna try here.

diogomartino (approved):
LGTM

PR #630, opened 2026-03-27
fix(626): Terminate active mic after leaveVoice()

### LINKED ISSUE #626: [Bug]: leaving a voice chat does not terminate active mic from browser
### Sharkord Version

0.0.15

### Environment

- **OS:** Linux
- **Browser:** Firefox & Chrome
- **Using reverse proxy:**nginx
- **Additional details:** Running in Docker

### Steps to Reproduce

1. Join Channel
2. Have fun
3. Leave channel
4. Observe Mic capture of browser

### Expected Behavior

Sharkord let's go of Mic once channel is left

### Actual Behavior

Mic capture stays until reload is hit

### Logs / Screenshots

Image

schemen:
## Summary
Currently, the mic remains active, showing the indicator while it should in theory release the resource. We teach the voice-provider to call cleanup() when the active voiceChannel becomes undefined.

Closes #626

### DIFF
--- a/apps/server/src/routers/voice/close-producer.ts
+++ b/apps/server/src/routers/voice/close-producer.ts
@@ -1,5 +1,6 @@
 import { Permission, ServerEvents, StreamKind } from '@sharkord/shared';
 import z from 'zod';
+import { logger } from '../../logger';
 import { VoiceRuntime } from '../../runtimes/voice';
 import { invariant } from '../../utils/invariant';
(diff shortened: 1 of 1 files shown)


schemen:
> I don't like that's another `useEffect` but fuck it, I really need to refactor that whole provider into something better. Right now it's getting to the point where it's not maintainable.

That's fair, let me know if there is a way to assist. I don't think I'm the right person to refactor it, but I can test a lot 👍

PR #636, opened 2026-03-28
fix(635): don't let users toggle camera/mic if already in progress

### LINKED ISSUE #635: [Bug]: My camera can appear off to me but everyone else can see me
### Sharkord Version

v0.16

### Environment

- **OS:** Fedora Linux 42
- **Browser:** Chrome 145, Firefox 147
- **Using reverse proxy:** nginx

### Steps to Reproduce

1. Get into a call with a friend
2. Toggle your camera on, then off right away

If you are unlucky, your friend will see your camera stream, even though your own interface shows your camera to be off.

### Expected Behavior

You should be able to click to toggle your camera on/off at will and expect that what you see is what your friend sees.  If your camera appears off to you, your friend should not be able to see you.

### Actual Behavior

If you are unlucky, the friend can see you even though your own interface says the camera is off.  The "unlucky" part is because it's a race condition so it's usually fine, but sometimes it's not.

### Logs / Screenshots

> Two users see different states

> Variation where one user sees a blank video feed and the other sees their camera off 

> Screenshare prompts can also pile up in Firefox from repeated clicking

dchester:
## Summary

In this PR we disregard repeated clicks on camera/mic/screenshare/deafen buttons if we're already in the process of toggling state.

Closes #635

## Additional Context

This addresses a very not-nice bug where a user can think their camera is off, but everyone else in the call can actually see them.  This also prevents piling up of "choose the screen to share" firefox dialogs if the user keeps pressing the screenshare button.

This all can happen in real life conditions, when an under-powered device is having trouble being responsive to UI events, and a frustrated user is repeatedly hitting the buttons in quick succession.

Some test cases:

- While in a voice channel we should be able to click *really fast* on the camera/mic/deafen toggles and everyone ends up seeing the same on/off state.

- In Firefox, we should be able to click repeatedly on "Start screen share" and see only one "Choose your screen" dialog.

- While not in a voice channel, we should be able to toggle mic and deafen states and see those settings honored when joining a voice channel.

### DIFF
--- a/apps/client/src/components/voice-provider/hooks/use-voice-controls.ts
+++ b/apps/client/src/components/voice-provider/hooks/use-voice-controls.ts
@@ -5,7 +5,7 @@ import { updateOwnVoiceState } from '@/features/server/voice/actions';
 import { useOwnVoiceState } from '@/features/server/voice/hooks';
 import { getTRPCClient } from '@/lib/trpc';
 import { getTrpcError } from '@sharkord/shared';
-import { useCallback } from 'react';
+import { useCallback, useRef } from 'react';
 import { toast } from 'sonner';
 
 type TUseVoiceControlsParams = {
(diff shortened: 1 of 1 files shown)


diogomartino (approved):
I'm not actually sure if there might be extra cases where that can happen, but this for sure will help a lot. Ty!

PR #649, opened 2026-04-04
feat(447): inline replies

### LINKED ISSUE #447: [Feature]: Replies should show as a regular message and just refer back to the original message
### Problem

Currently replies are implemented as threads, but that breaks the flow that normal replies generally afford.

### Proposed Solution

Reply's should be like normal messages with a smaller, truncated version of the original message above it. Clicking that could jump back to the position of the original message

diogomartino:
## Summary

Closes #447

## Additional Context

image


SSTHacksword:
This is awesome!

PR #659, opened 2026-04-05
feat(62): add fullscreen button to video/screen share cards

### LINKED ISSUE #62: [request] full screen mode
It would be nice to have a full‑screen option (browser independent, so to speak) while spectating 😀

jagobainda:
Adds a fullscreen button to screen share, external stream, and external video cards in voice channels ( Closes #62 ). When clicked, the card enters native browser fullscreen and the zoom state is reset. Exiting fullscreen restores the card to its original layout, removing borders and rounded corners while in fullscreen mode.

The button is positioned at the bottom-right of the card, following the same hover-reveal pattern used by the pin button. If a different placement or behavior is preferred, feel free to leave a comment.

image


diogomartino:
I would put the full screen on the top right near the other buttons tbh.

jagobainda:
> I would put the full screen on the top right near the other buttons tbh.

Done.

Also done: added fullscreen idle behavior and shortcuts.

- Controls and info bar now auto-hide after 3 seconds of inactivity while in fullscreen, and reappear on mouse movement, touch, or key press. The cursor is also hidden during idle.
- Double-click on the video area toggles fullscreen.
- Pressing "F" while in fullscreen exits it.

Same, if you have any preference regarding any behavior or feature, let me know.

jagobainda:
Resolved a minor conflict. I was aware that external-video-card.tsx was dead code, but I had added fullscreen support to it just in case.

diogomartino:
Why F? Esc would be better IMO

jagobainda:
> Why F? Esc would be better IMO

I’ve chosen F because it’s the key used by major streaming services, but ESC also works since it’s the default for native fullscreen in the browsers that the feature uses

diogomartino (approved):
Looks good, ty!

PR #662, opened 2026-04-07
feat(628): Fast message deletion with Shift + click on trash icon

### LINKED ISSUE #628: [Feature]: Fast message deletion with Shift + click on trash icon
### Problem

Deleting messages currently always opens a confirmation dialog for each individual message, which makes cleaning up or moderating multiple messages slow and repetitive. This is especially noticeable for moderators or power users who need to delete spam or obvious unwanted content in quick succession.

### Proposed Solution

Introduce a shortcut where holding **Shift** while clicking the delete (trash) icon immediately deletes the selected message without showing the confirmation dialog. The existing behavior (showing a confirmation dialog) should remain unchanged when clicking the delete icon without **Shift**, to keep deletions safe for regular users. Optionally, add a user or role-based setting such as “Enable fast deletion with Shift + click” so that only users who explicitly opt in (for example, moderators with message management permissions) can use this faster workflow.

The-Blind-Magician:
## Summary
Adds an option to "super delete" messages by Shift+Click-ing the message delete button. (Closes #628)

The original hover behavior is intact for MessageActions. However, now if Shift is held while hovering, a new Trash icon is displayed. Clicking the Trash icon while shift is held will trigger a message deletion with no confirmation window. 

Hover with Shift
image

After Shift+Click
image

### DIFF
--- a/apps/client/src/components/channel-view/text/message-actions.tsx
+++ b/apps/client/src/components/channel-view/text/message-actions.tsx
@@ -6,6 +6,7 @@ import {
   type TEmojiItem
 } from '@/components/tiptap-input/helpers';
 import { openThreadSidebar } from '@/features/app/actions';
+import { useIsShiftHeld } from '@/features/app/hooks';
 import { requestConfirmation } from '@/features/dialogs/actions';
 import { getTRPCClient } from '@/lib/trpc';
 import { Permission } from '@sharkord/shared';
@@ -26,9 +27,7 @@ import {
   forwardRef,
   memo,
   useCallback,
-  useEffect,
-  useMemo,
-  useState
+  useMemo
 } from 'react';
 import { useTranslation } from 'react-i18next';
 import { toast } from 'sonner';
@@ -48,7 +47,6 @@ type TMessageActionsProps = {
   isThreadReply?: boolean;
   isPinned?: boolean;
   disablePin?: boolean;
-  isHovered?: boolean;
 };
 
 const MessageActions = memo(
@@ -61,8 +59,7 @@ const MessageActions = memo(
     isThreadReply,
     isPinned,
     disablePin,
-    onReply,
-    isHovered
+    onReply
   }: TMessageActionsProps) => {
     const { t } = useTranslation();
     const { recentEmojis } = useRecentEmojis();
@@ -71,35 +68,10 @@ const MessageActions = memo(
       [recentEmojis]
     );
 
-    const [isShiftDown, setIsShiftDown] = useState(false);
-
-    useEffect(() => {
-      if (!isHovered) return;
-
-      setIsShiftDown(false);
-
-      const onKeyDown = (e: KeyboardEvent) => {
-        if (e.key === 'Shift') {
-          setIsShiftDown(true);
-        }
-      };
-      const onKeyUp = (e: KeyboardEvent) => {
-        if (e.key === 'Shift') {
-          setIsShiftDown(false);
-        }
-      };
-
-      document.addEventListener('keydown', onKeyDown);
-      document.addEventListener('keyup', onKeyUp);
-
-      return () => {
-        document.removeEventListener('keydown', onKeyDown);
-        document.removeEventListener('keyup', onKeyUp);
-      };
-    }, [isHovered]);
+    const isShiftHeld = useIsShiftHeld();
 
     const onDeleteClick = useCallback(async () => {
-      if (!isShiftDown) {
+      if (!isShiftHeld) {
         const choice = await requestConfirmation({
           title: t('deleteMessageTitle'),
           message: t('deleteMessageConfirm'),
@@ -117,7 +89,7 @@ const MessageActions = memo(
       } catch {
         toast.error(t('failedDeleteMessage'));
       }
-    }, [isShiftDown, messageId, t]);
+    }, [isShiftHeld, messageId, t]);
 
     const onEmojiSelect = useCallback(
       async (emoji: TEmojiItem) => {
@@ -189,7 +161,7 @@ const MessageActions = memo(
             <IconButton
               size="sm"
               variant="ghost"
-              icon={isShiftDown ? TRASH2_RED : Trash}
+              icon={isShiftHeld ? TRASH2_RED : Trash}
               onClick={onDeleteClick}
               title={t('deleteMessageTitle')}
             />
--- a/apps/client/src/components/channel-view/text/message-actions.tsx
+++ b/apps/client/src/components/channel-view/text/message-actions.tsx
@@ -19,18 +19,13 @@ import {
   Reply,
   Smile,
   Trash,
-  Trash2,
-  type LucideIcon,
-  type LucideProps
+  Trash2
 } from 'lucide-react';
-import { forwardRef, memo, useCallback, useMemo } from 'react';
+import { memo, useCallback, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import { toast } from 'sonner';
 
 const MAX_QUICK_EMOJIS = 4;
-const TRASH2_RED: LucideIcon = forwardRef<SVGSVGElement, LucideProps>(
-  (props, ref) => <Trash2 ref={ref} {...props} className={'text-destructive'} />
-);
 
 type TMessageActionsProps = {
   messageId: number;
@@ -156,7 +151,8 @@ const MessageActions = memo(
             <IconButton
               size="sm"
               variant="ghost"
-              icon={isShiftHeld ? TRASH2_RED : Trash}
+              icon={isShiftHeld ? Trash2 : Trash}
+              className={isShiftHeld ? 'text-destructive' : ''}
               onClick={onDeleteClick}
               title={t('deleteMessageTitle')}
             />
--- a/apps/client/src/components/channel-view/text/message-actions.tsx
+++ b/apps/client/src/components/channel-view/text/message-actions.tsx
@@ -76,6 +76,8 @@ const MessageActions = memo(
     useEffect(() => {
       if (!isHovered) return;
 
+      setIsShiftDown(false);
+
       const onKeyDown = (e: KeyboardEvent) => {
         if (e.key === 'Shift') {
           setIsShiftDown(true);
--- a/apps/client/src/components/channel-view/text/message-actions.tsx
+++ b/apps/client/src/components/channel-view/text/message-actions.tsx
@@ -35,7 +35,7 @@ import { toast } from 'sonner';
 
 const MAX_QUICK_EMOJIS = 4;
 const TRASH2_RED: LucideIcon = forwardRef<SVGSVGElement, LucideProps>(
-  (props, ref) => <Trash2 ref={ref} {...props} color="red" />
+  (props, ref) => <Trash2 ref={ref} {...props} className={'text-destructive'} />
 );
 
 type TMessageActionsProps = {
--- a/apps/client/src/components/channel-view/text/message-actions.tsx
+++ b/apps/client/src/components/channel-view/text/message-actions.tsx
@@ -22,14 +22,21 @@ import {
   type LucideIcon,
   type LucideProps
 } from 'lucide-react';
-import { forwardRef, memo, useCallback, useEffect, useMemo, useState } from 'react';
+import {
+  forwardRef,
+  memo,
+  useCallback,
+  useEffect,
+  useMemo,
+  useState
+} from 'react';
 import { useTranslation } from 'react-i18next';
 import { toast } from 'sonner';
 
 const MAX_QUICK_EMOJIS = 4;
-const TRASH2_RED: LucideIcon = forwardRef<SVGSVGElement, LucideProps>((props, ref) => (
-  <Trash2 ref={ref} {...props} color="red" />
-));
+const TRASH2_RED: LucideIcon = forwardRef<SVGSVGElement, LucideProps>(
+  (props, ref) => <Trash2 ref={ref} {...props} color="red" />
+);
 
 type TMessageActionsProps = {
   messageId: number;
@@ -105,7 +112,7 @@ const MessageActions = memo(
       } catch {
         toast.error(t('failedDeleteMessage'));
       }
-    }, [isShiftDown,messageId, t]);
+    }, [isShiftDown, messageId, t]);
 
     const onEmojiSelect = useCallback(
       async (emoji: TEmojiItem) => {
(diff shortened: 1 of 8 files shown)


apps/client/src/components/channel-view/text/message-actions.tsx
```diff
@@ -56,25 +62,50 @@ const MessageActions = memo(
...
+      };
+
+      document.addEventListener('keydown', onKeyDown);
+      document.addEventListener('keyup', onKeyUp);
+
+      return () => {
+        document.removeEventListener('keydown', onKeyDown);
+        document.removeEventListener('keyup', onKeyUp);
+      };
+    }, []);
```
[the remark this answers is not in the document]
The-Blind-Magician: Addressed in commit a5b215c

apps/client/src/components/channel-view/text/message-actions.tsx
```diff
@@ -56,25 +62,50 @@ const MessageActions = memo(
...
+
+      const trpc = getTRPCClient();
       try {
         await trpc.messages.delete.mutate({ messageId });
         toast.success(t('messageDeleted'));
       } catch {
         toast.error(t('failedDeleteMessage'));
       }
-    }, [messageId, t]);
+    }, [isShiftDown,messageId, t]);
```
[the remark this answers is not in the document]
The-Blind-Magician: Fixed in commit 06e5ad7

apps/client/src/components/channel-view/text/message-actions.tsx
```diff
@@ -17,13 +17,19 @@ import {
...
 } from 'lucide-react';
-import { memo, useCallback, useMemo } from 'react';
+import { forwardRef, memo, useCallback, useEffect, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { toast } from 'sonner';
 
 const MAX_QUICK_EMOJIS = 4;
+const TRASH2_RED: LucideIcon = forwardRef<SVGSVGElement, LucideProps>((props, ref) => (
+  <Trash2 ref={ref} {...props} color="red" />
+));
```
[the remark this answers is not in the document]
The-Blind-Magician: Fixed in commit 8003d57

apps/client/src/components/channel-view/text/message-actions.tsx
```diff
@@ -17,13 +17,26 @@ import {
...
+  useMemo,
+  useState
+} from 'react';
 import { useTranslation } from 'react-i18next';
 import { toast } from 'sonner';
 
 const MAX_QUICK_EMOJIS = 4;
+const TRASH2_RED: LucideIcon = forwardRef<SVGSVGElement, LucideProps>(
+  (props, ref) => <Trash2 ref={ref} {...props} className={'text-destructive'} />
+);
```
diogomartino: What is this? Doesn't make sense.

apps/client/src/components/channel-view/text/message-actions.tsx
```diff
@@ -56,25 +71,53 @@ const MessageActions = memo(
...
+        }
+      };
+      const onKeyUp = (e: KeyboardEvent) => {
+        if (e.key === 'Shift') {
+          setIsShiftDown(false);
+        }
+      };
 
+      document.addEventListener('keydown', onKeyDown);
+      document.addEventListener('keyup', onKeyUp);
```
diogomartino: Adding event listeners here might not be a good idea. If there are 200 messages being rendered, this will register 200 event listeners. I would say move this logic to a parent but that will get us in prop drilling hell. Best way is to save the state of modifier keys directly in the store in the app slice and in `HotkeysController` create the logic to sync the state. Then create the selector and hooks like all the other hooks and selectors are created.

apps/client/src/components/channel-view/text/message-actions.tsx
```diff
@@ -146,7 +189,7 @@ const MessageActions = memo(
             <IconButton
               size="sm"
               variant="ghost"
-              icon={Trash}
+              icon={isShiftDown ? TRASH2_RED : Trash}
```
diogomartino: You can pass the <Trash2 className=.../> here directly I think.

diogomartino (changes requested):

apps/client/src/components/channel-view/text/message-actions.tsx
The-Blind-Magician: The way I currently have it implemented, the parent element passes an isHovered prop which useEffect has as a dependency. If the parent Message is currently being hovered over, the useEffect will register the listeners. If that changes, the useEffect deconstructor will remove the listeners before reevaluation. Since the isHovered prop now evaluates to false, the useEffect early returns without creating the listeners. Since only 1 Message can be hovered over at a time, only 1 keyUp and 1 keyDown listener will be active at a time. 

I can certainly check into using the HotkeysController if desired. I wasn't aware that existed when throwing this together.

apps/client/src/components/channel-view/text/message-actions.tsx
The-Blind-Magician: This was my workaround in order to use a red Trash2 icon without having to modify the IconButton class. Through testing, it didn't seem that changing between White icons was flashy enough to draw the user's attention. Red seemed more apt. 

Unfortunately, the icon is passed as a prop to the child as an uninstantiated object without any modifiers. To get around modifying IconButton (to avoid unintended trickle down effects), I use forwardRef to set up the object without overriding IconButton class' ability to control its instantiation (for future proofing). Setting className="text-destructive" was a way to apply a red coloring to the icon while respecting the application's overall Tailwind "theme".

Perhaps there's a better way to do this that I am not aware of.

apps/client/src/components/channel-view/text/message-actions.tsx
The-Blind-Magician: I don't quite understand your comment. I am using the ternary to pass either a red Trash2 icon or a white Trash icon. Basing the ternary on a useState variable will ensure we trigger a re-render when isShiftDown changes.

apps/client/src/components/channel-view/text/message-actions.tsx
The-Blind-Magician: Re-implemented with HotkeysController in commit 35e78f0. I went ahead and added Ctrl and Alt as usable hotkeys as well. They are stored collectively for extensibility but have individual selectors. 

There is one small but notable downside to this approach. Since keys are captured at the application level (instead of component level), if a user is holding a hotkey while they click out of the window, the keyUp event won't be captured correctly. When done at the component level, this could be avoided by defaulting the value upon re-render. As far as I am aware, we have no such option in this case. As such, the hotkey will still be registered as "pressed" when focus is returned to the main window.

I've addressed this by adding a "blur" handler to the HotkeysController to reset the hotkey states. I don't prefer "crossing wires" like this, but it does solve the problem.

apps/client/src/components/channel-view/text/message-actions.tsx
diogomartino: I understood why you did it, but there's a much easier way. See below.

apps/client/src/components/channel-view/text/message-actions.tsx
diogomartino: From what I've seen, you want to turn the icon red when shift is pressed, right? So you can just pass a className to <IconButton> with the color and it's gonna apply to whatever icon is there, so you can do something like:

            <IconButton
              className={isShiftPressed ?? 'text-red-500'}
            />

apps/client/src/components/channel-view/text/message-actions.tsx
diogomartino: Sorry, didn't notice that part of `isHovered`. In that case that would've been fine. But the implementation in the store might be useful for other features so it's ok.

apps/client/src/features/app/actions.ts
```diff
@@ -251,3 +252,8 @@ export const togglePluginSlotDebug = () => {
   store.dispatch(appSliceActions.setPluginSlotDebug(next));
   setLocalStorageItemBool(LocalStorageKey.PLUGIN_SLOT_DEBUG, next);
 };
+
+export const setHotkeyIsHeld = (keysDown: Record<string, boolean>) => {
+  store.dispatch(appSliceActions.setHotkeyIsHeld(keysDown));
+  setLocalStorageItemAsJSON(LocalStorageKey.HOTKEY_IS_HELD, keysDown);
```
diogomartino: Doesn't make sense to persist this in the local storage.

apps/client/src/features/app/slice.ts
```diff
@@ -70,6 +72,10 @@ const initialState: TAppState = {
   pluginSlotDebug: getLocalStorageItemBool(
     LocalStorageKey.PLUGIN_SLOT_DEBUG,
     false
+  ),
+  hotkeyIsHeld: getLocalStorageItemAsJSON<Record<string, boolean>>(
```
diogomartino: Same as above.

apps/client/src/components/hotkeys-controller/index.tsx
```diff
@@ -1,19 +1,49 @@
...
 const HotkeysController = memo(() => {
   const handleKeyDown = useCallback((e: KeyboardEvent) => {
     if (e.key === 'F4') {
       togglePluginSlotDebug();
     }
+
+    const hotkeyState: Record<string, boolean> = {};
+    hotkeyState['Shift'] = e.shiftKey;
+    hotkeyState['Control'] = e.ctrlKey;
+    if (e.key === 'Alt') {
```
diogomartino, line 13: Why is this needed?

apps/client/src/components/hotkeys-controller/index.tsx
```diff
@@ -1,19 +1,49 @@
...
+    hotkeyState['Alt'] = e.altKey;
+    setHotkeyIsHeld(hotkeyState);
+  }, []);
+
+  const handleKeyUp = useCallback((e: KeyboardEvent) => {
+    const hotkeyState: Record<string, boolean> = {};
+    hotkeyState['Shift'] = e.shiftKey;
+    hotkeyState['Control'] = e.ctrlKey;
+    hotkeyState['Alt'] = e.altKey;
+    setHotkeyIsHeld(hotkeyState);
```
diogomartino: IMO this would be cleaner just with:

```
setHotkeyIsHeld({
Shift: e.shiftKey,
...etc
})
```

Right now it's too verbose.

apps/client/src/features/app/selectors.ts
```diff
@@ -72,3 +72,21 @@ export const threadSidebarDataSelector = createSelector(
     channelId
   })
 );
+
+export const isShiftHeldSelector = (state: IRootState) => {
```
diogomartino: These all can be turned into one liners:

`export const isShiftHeldSelector = (state: IRootState) => state.app.hotkeyIsHeld?.Shift ?? false`

apps/client/src/features/app/slice.ts
```diff
@@ -26,6 +27,7 @@ export interface TAppState {
   voiceChatSidebarOpen: boolean;
   voiceChatChannelId: number | undefined;
   pluginSlotDebug: boolean;
+  hotkeyIsHeld: Record<string, boolean> | undefined;
```
diogomartino: This should never be undefined, just initialize it as an object with all keys false. Also, the naming is off. "modifierKeysHeldMap" is clearer.

diogomartino (changes requested):
Just small things!

apps/client/src/components/channel-view/text/message-actions.tsx
The-Blind-Magician: It didn't occur to me that the property would trickle down. Fixed in 15a9940. I did have to use a ternary instead of nullish since isShiftPressed will always be defined. I've applied the same logic to the icon so we can have a "loud" visual change when active.

apps/client/src/components/channel-view/text/message-actions.tsx
The-Blind-Magician: Ah, I see. Addressed below.

apps/client/src/features/app/actions.ts
The-Blind-Magician: The difference between local storage and session storage was lost on my 2am self. Fixed in 0d2e5ff.

apps/client/src/features/app/slice.ts
The-Blind-Magician: Fixed in 0d2e5ff.

apps/client/src/features/app/selectors.ts
The-Blind-Magician: Fixed in 0d2e5ff.

apps/client/src/features/app/slice.ts
The-Blind-Magician: Fixed in 0d2e5ff.

apps/client/src/components/hotkeys-controller/index.tsx
The-Blind-Magician: Allowing Alt to behave normally could/would allow for the browser's Alt+Key shortcuts to take precedence over Sharkord. During testing I found that it would forcibly focus the browser itself, un-focusing the app. If we want to maintain the user's ability to do so, I can remove the preventDefault.

apps/client/src/components/hotkeys-controller/index.tsx
The-Blind-Magician: Fixed in 7eca3cf.

apps/client/src/components/hotkeys-controller/index.tsx
diogomartino: That makes sense, ye. You can leave it

apps/client/src/components/channel-view/text/message-actions.tsx
```diff
@@ -146,7 +151,8 @@ const MessageActions = memo(
             <IconButton
               size="sm"
               variant="ghost"
-              icon={Trash}
+              icon={isShiftHeld ? Trash2 : Trash}
```
diogomartino, line 154: Why change the icon itself when shift is pressed?

diogomartino (changes requested):
Looks good, just the icon change is a bit awkward IMO, but it's fine. Thanks for all the changes!

diogomartino (approved):

apps/client/src/components/channel-view/text/message-actions.tsx
The-Blind-Magician: The "Trash" icon is a hollow icon, so the change in color is less noticable. "Trash2" has some infill, so it makes the color change pop a bit more. Personally, I prefer the look of "Trash2" in either case but wanted to respect the original design choice.

PR #680, opened 2026-04-11
feat(678): Mute Hotkey not working + framework for adjustable hotkey configuration 

### LINKED ISSUE #678: [Feature]: Mute Hotkey not working + adjustable hotkey configuration
### Problem

Mute Hotkey (Ctrl+shift+M) doesnt work; changing the hotkeys for specific commands would also be pretty cool (or straight up creating your own commands)

### Proposed Solution

adjustable configuration of the hotkeys/commands in user settings.

The-Blind-Magician:
## Summary
This PR adds a new client-side structure called a ShortcutRegistrar. This was an offshoot of the HotkeysController, aimed at providing function callback behavior for dynamically registered keyboard shortcuts. No more writing keyDown listeners. In the spirit of the original issue, I've also implemented shortcuts for Mute and Deafen, as per their description in the sidebar.json.
Closes #678 

## Additional Context
I've left the "setModifierKeysHeldMap" behavior intact to allow for more dynamic application of the modifier keys in cases where a shortcut is not explicitly needed (e.g. shift for super-delete). 

While the framework is in place and preliminary function seems promising, I didn't feel like it was my place to decide what actions do and don't warrant user-configurable shortcut keys. As such, I have not implemented anything past the framework for registering those custom shortcuts. If desired I can throw something together, or we can put that up for discussion in it's own Issue.

### DIFF
--- a/apps/client/src/components/hotkeys-controller/index.tsx
+++ b/apps/client/src/components/hotkeys-controller/index.tsx
@@ -6,39 +6,39 @@ import { memo, useEffect } from 'react';
 import { ShortcutRegistrar } from '../shortcut-registrar';
 
 const HotkeysController = memo(() => {
-  const pressedKeys = new Set<string>();
-
-  const handleKeyDown = (e: KeyboardEvent) => {
-    if (e.repeat) return;
-    pressedKeys.add(e.key);
-    ShortcutRegistrar.submit(pressedKeys, e);
-
-    setModifierKeysHeldMap({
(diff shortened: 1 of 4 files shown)


diogomartino:
Sorry, this is not how I want this implemented. The original issue doesn't even make sense since Sharkord doesn't have hotkeys. So a thing that doesn't exist cannot not work :laughing:.

I will take care of hotkeys when the time comes and the implementation will depend on how the desktop app will work.

PR #704, opened 2026-04-24
fix(703): don't let too many files overflow chat input max height

### LINKED ISSUE #703: [Bug]: Resizing chat input is jumpy with lots of files on small screens
### Sharkord Version

91b2cefe7ab0b29b10ad49dd19d98ec354b70c61

### Environment

- **OS:** Linux
- **Browser:** Chrome, Firefox

### Steps to Reproduce

1. Size your browser to ~900px wide
2. Upload 6+ image files
3. Observe the chat input is off the bottom of the screen
4. Also observe that resizing jumps up while dragging

### Expected Behavior

- The message composition section should expand to fit all the files and at least one line of text in the chat input.  

- If there are so many files that they simply cannot fit (within the absolute max of `80vh`) only then should the chat input be allowed to flow off the bottom of the screen

- Resizing up and down with many files should be smooth even on a small screen

### Actual Behavior

If the height of the files section is larger than the default max height of the message composition area, then the section may not expand to the height necessary to keep the chat input in view.

### Logs / Screenshots

https://github.com/user-attachments/assets/532bca9e-ceea-4d69-8cec-cc54d8fc41ab

dchester:
## Summary

In this PR we deal better with resizing the `message-compose` component when there are many files on a small screen. 

Closes #703

## Additional Context

This addresses two issues...

#### 1) Files causing container to grow beyond default max height

When we add so many files that it would make `message-compose` grow taller than the default max height of the component, the measurement of the "natural height" was inadvertently being clamped to that max height, resulting in jumpy resizing behavior.

To address this, we unset `max-height` before we measure, and restore it afterwards.

https://github.com/user-attachments/assets/e29239b3-a22a-4e80-8941-ba26d64e0d37
> When files cause the container to grow beyond the default max height, drag-drop resizing still works smoothly

#### 2) Even more files causing container to grow beyond viewport height

With even more files, they could make the container be taller than the entire viewport, making some content unreachable.

To address this, we clamp the container height to an absolute max height of `80vh`, and then restore the previous height (if you had dragged to an explicit height), or else the max height (if you had not) when the files are either removed, or sent with the message.

### DIFF
--- a/apps/client/src/components/channel-view/text/chat-input-divider.tsx
+++ b/apps/client/src/components/channel-view/text/chat-input-divider.tsx
@@ -44,7 +44,7 @@ const ChatInputDivider = ({
       const startHeightPx = getHeight(composeEl);
 
       const maxPx = (MAX_VH / 100) * window.innerHeight;
-      const minPx = measureMinHeight(composeEl);
+      const minPx = Math.min(measureMinHeight(composeEl), maxPx);
 
       composeEl.style.maxHeight = '';
       composeEl.style.height = `${startHeightPx}px`;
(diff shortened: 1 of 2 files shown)


dchester:
For the record, this is a fix for the issue mentioned in https://github.com/Sharkord/sharkord/pull/535#issuecomment-4270257052

diogomartino (approved):
LGTM, ty!

PR #714, opened 2026-04-27
fix(713): update storage quota checks in getExceedingOldFiles function

### LINKED ISSUE #713: [Bug]: DELETE_OLD_FILES overflow mode uses per-file size limit instead of total storage quota
### Sharkord Version

v0.0.19

### Environment

- **OS:** does not apply
- **Browser:** does not apply
- **Using reverse proxy:** does not apply
- **Additional details:** does not apply

### Steps to Reproduce

1. Configure `storageQuota` smaller than the sum of expected uploads, with `storageUploadMaxFileSize` set to the typical per-file value (e.g. quota = 10 GB, max file size = 2 GB).
2. Set `storageOverflowAction = DELETE_OLD_FILES`.
3. Fill the server up to or near `storageQuota`.
4. Upload a new file.

### Expected Behavior

When `storageOverflowAction` is set to `DELETE_OLD_FILES` and a new upload would push total storage over `storageQuota`, the oldest files are deleted until enough space is freed to accommodate the new file. Total server storage stays at or below `storageQuota`.

### Actual Behavior

`getExceedingOldFiles` computes the overflow against `storageUploadMaxFileSize` (per-file limit) instead of `storageQuota` (total server cap), producing an incorrect amount of space to free:

- When `storageUploadMaxFileSize` < `storageQuota` (typical config): the calculated overflow is inflated, so the function deletes far more files than necessary.
- When `storageUploadMaxFileSize` > `storageQuota` (inverted config): the calculated overflow is negative or zero, so the function returns an empty list and skips deletion entirely — the server silently exceeds its quota without any pruning.

### Logs / Screenshots

_No response_

jagobainda:
## Fix: use `storageQuota` to compute storage overflow in `getExceedingOldFiles`

Closes #713 

## What changed

In `apps/server/src/db/queries/files.ts`, `getExceedingOldFiles` now calculates how much space must be freed against `storageQuota` (total server cap) instead of `storageUploadMaxFileSize` (per-file cap). The per-file pre-check is kept and its error message is corrected.

### Before

```ts
const { storageUploadMaxFileSize } = await getSettings();

if (newFileSize > storageUploadMaxFileSize) {
  throw new Error('File size exceeds total server storage quota');
}

const currentTotalSize = Number(currentUsage?.totalSize ?? 0);
const wouldExceedBy =
  currentTotalSize + newFileSize - storageUploadMaxFileSize;
```

### After

```ts
const { storageQuota, storageUploadMaxFileSize } = await getSettings();

if (newFileSize > storageUploadMaxFileSize) {
  throw new Error('File size exceeds the maximum allowed file size');
}

const currentTotalSize = Number(currentUsage?.totalSize ?? 0);
const wouldExceedBy = currentTotalSize + newFileSize - storageQuota;
```

## Why

`getExceedingOldFiles` is invoked from `file-manager.ts` only after the outer check `newServerStorage > storageQuota` has already determined that the **aggregate** server quota would be exceeded. Its job is to decide how many old files to evict to bring the aggregate back under that quota, so the math must reference `storageQuota`, not the per-file limit.

The two responsibilities are now cleanly separated:

- `storageUploadMaxFileSize` is used **only** for the per-file pre-check (rejecting an upload that would be invalid even with unlimited space).
- `storageQuota` is used for the aggregate overflow calculation that drives which files get pruned.

The pre-check error message is also corrected: it previously claimed "total server storage quota" while the condition actually compares against the per-file limit.

## Effect on `DELETE_OLD_FILES`

With the previous formula, `wouldExceedBy` was computed against `storageUploadMaxFileSize` instead of `storageQuota`, producing an incorrect amount of space to free:

- When `storageUploadMaxFileSize` < `storageQuota` (typical config): `wouldExceedBy` was inflated, causing far more files to be deleted than necessary.
- When `storageUploadMaxFileSize` > `storageQuota` (inverted config): `wouldExceedBy` was negative or zero, so the function returned `[]` and no files were ever pruned — the server silently exceeded its quota.

After this fix, `DELETE_OLD_FILES` correctly evicts only the oldest files needed to bring aggregate usage back within `storageQuota`.

## Risk / compatibility

- No schema changes, no settings changes, no API surface changes.
- The only observable behavior change occurs when `DELETE_OLD_FILES` is configured: it now actually deletes files (which is the intended behavior).
- The error message change is user-facing only when an upload exceeds `storageUploadMaxFileSize`; the new wording matches what the check actually verifies.

## Suggested follow-ups (out of scope)

- Add a unit test for `getExceedingOldFiles` covering the case `storageUploadMaxFileSize < storageQuota` to prevent regressions.

### DIFF
--- a/apps/server/src/db/queries/files.ts
+++ b/apps/server/src/db/queries/files.ts
@@ -6,10 +6,10 @@ import { files, messageFiles } from '../schema';
 import { getSettings } from './server';
 
 const getExceedingOldFiles = async (newFileSize: number) => {
-  const { storageUploadMaxFileSize } = await getSettings();
+  const { storageQuota, storageUploadMaxFileSize } = await getSettings();
 
   if (newFileSize > storageUploadMaxFileSize) {
-    throw new Error('File size exceeds total server storage quota');
+    throw new Error('File size exceeds the maximum allowed file size');
   }
 
   const currentUsage = await db
@@ -20,8 +20,7 @@ const getExceedingOldFiles = async (newFileSize: number) => {
     .get();
 
   const currentTotalSize = Number(currentUsage?.totalSize ?? 0);
-  const wouldExceedBy =
-    currentTotalSize + newFileSize - storageUploadMaxFileSize;
+  const wouldExceedBy = currentTotalSize + newFileSize - storageQuota;
 
   if (wouldExceedBy <= 0) {
     return [];


jagobainda:
I can add the test that covers this case if needed

diogomartino (approved):
LGTM

PR #720, opened 2026-05-05
fix: post-sanitization guard checks original input instead of sanitized content

jagobainda:
Small logic bug: the second `invariant` in both `send-message.ts` and `edit-message.ts`
reads `input.content` instead of the sanitized variable, so content that sanitizes
to an empty string (e.g. `<script>`) bypasses the guard and gets stored as `""`.

No real attack surface — requires authentication + SEND_MESSAGES permission, and the
XSS payload itself is never stored. Pure data integrity issue.

Not worth a separate issue given the triviality of the fix.

**Changed:** `isEmptyMessage(input.content)` → `isEmptyMessage(targetContent / sanitizedContent)` in both files.


diogomartino (approved):
LGTM

PR #723, opened 2026-05-10
fix(722): add rate limiting to unprotected high-cost endpoints

### LINKED ISSUE #722: [Bug]: several high-frequency / high-cost endpoints lack server-side rate limiting (potential DoS surface)
### Sharkord Version

v0.0.19

### Environment

- **OS:** does not apply
- **Browser:** does not apply
- **Using reverse proxy:** does not apply
- **Additional details:** does not apply

A subset of high-frequency or high-cost tRPC endpoints expose `protectedProcedure` (and one `publicProcedure`) directly, with no server-side rate limiter applied. Each call performs multiple SQLite queries (some with JOINs, some unbounded reads) against a single-writer database. Any authenticated client connecting directly over WebSocket can bypass whatever client-side throttling exists and sustain hundreds of calls per second per endpoint, generating thousands of DB queries per second.

### Steps to Reproduce

1. Authenticate against the WebSocket server with a valid token (or, for the public endpoint, none at all).
2. Connect directly over WebSocket and invoke any of the affected `messages.*`, `channels.*`, `dms.*`, `emojis.*`, or `others.*` mutations/queries in a tight loop, ignoring the official client's UI flow.
3. Observe the server executing the full permission stack and underlying queries on every single call, with no `TOO_MANY_REQUESTS` response. 

### Expected Behavior

All endpoints whose cost per call is non-trivial / multiple queries, JOINs, unbounded reads, file I/O, or input-driven amplification — are rate-limited server-side, returning `TOO_MANY_REQUESTS` once a reasonable per-client budget is exceeded. This is consistent with the protection already applied to `sendMessage`, `searchMessages`, `joinVoice`, and `joinServer`.

### Actual Behavior

Several such endpoints are reachable without any server-side limit

### Logs / Screenshots

_No response_

jagobainda:
### Problem
Several tRPC endpoints had no server-side rate limit, making them viable DoS vectors. Any authenticated client connecting directly over WebSocket could bypass the official client's UI flow and call them in a tight loop, generating thousands of SQLite queries per second against a single-writer database.

### Fix
Wrap each affected endpoint with the existing `rateLimitedProcedure` infrastructure. All limits are added to `config.rateLimiters.*` so they can be tuned via `config.ini` without touching code.

| Endpoint | Limit (per IP) | Rationale |
|---|---|---|
| `messages.signalTyping` | 40 / 5 s | Legitimate max at `TYPING_MS = 300 ms` is ~16/5 s. Cap gives ~2.5× headroom and accounts for multiple typists behind a shared NAT. |
| `messages.getMessages` | 60 / 10 s | Heaviest read endpoint (~7 queries + JOINs + pubsub). Cap comfortably covers rapid channel-switching and scroll pagination. |
| `messages.toggleMessageReaction` | 60 / 10 s | 4 queries per call; same threat profile as `signalTyping`. |
| `channels.markAsRead` | 60 / 10 s | 3 queries per call; triggered naturally by channel-switching. |
| `dms.openDirectMessage` | 10 / 60 s | Creates persistent state + publishes permission events; called at most a few times per session legitimately. |
| `emojis.addEmoji` | 10 / 60 s | Loops on `fileManager.saveFile` + DB insert per array element with no upper bound; admin-gated but still unbounded. |
| `others.handshake` | 10 / 60 s | Only `publicProcedure` — reachable pre-auth with no prior limit. |

The chosen rates are a starting point and may need tuning in real deployments, particularly for the navigation endpoints on servers where many users share an IP. Treat them as debatable.

Some unprotected endpoints were intentionally left out: those gated behind admin permissions (limited blast radius), low-cost single-lookup queries (not worth the noise), and endpoints whose cost lives inside a delegated subsystem where the right place to throttle isn't the tRPC layer.

Closes #722

### DIFF
--- a/apps/server/src/config.ts
+++ b/apps/server/src/config.ts
@@ -41,6 +41,34 @@ const zConfig = z.object({
     search: z.object({
       maxRequests: z.coerce.number().int().positive(),
       windowMs: z.coerce.number().int().positive()
+    }),
+    signalTyping: z.object({
+      maxRequests: z.coerce.number().int().positive(),
+      windowMs: z.coerce.number().int().positive()
+    }),
+    getMessages: z.object({
+      maxRequests: z.coerce.number().int().positive(),
+      windowMs: z.coerce.number().int().positive()
+    }),
+    markAsRead: z.object({
+      maxRequests: z.coerce.number().int().positive(),
+      windowMs: z.coerce.number().int().positive()
+    }),
+    toggleMessageReaction: z.object({
+      maxRequests: z.coerce.number().int().positive(),
+      windowMs: z.coerce.number().int().positive()
+    }),
+    addEmoji: z.object({
+      maxRequests: z.coerce.number().int().positive(),
+      windowMs: z.coerce.number().int().positive()
+    }),
+    openDirectMessage: z.object({
+      maxRequests: z.coerce.number().int().positive(),
+      windowMs: z.coerce.number().int().positive()
+    }),
+    handshake: z.object({
+      maxRequests: z.coerce.number().int().positive(),
+      windowMs: z.coerce.number().int().positive()
     })
   })
 });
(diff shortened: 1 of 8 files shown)


diogomartino (approved):
I'll approve this but probably in the future the rate limiting config will be a separate file so we don't pollute the main config that much.

PR #750, opened 2026-05-30
feat(28): drag and drop users between voice channels

### LINKED ISSUE #28: Move users between voice channels
Drag and drop users between channels

### LINKED ISSUE #745: [Feature]: Add more moderation tools on users in voice channels
### Problem

It would be great if users with suitable permissions would be able to move users between voice channels and kick them from a voice channel by right clicking them in the voice channel

Moving can also be done by dragging and dropping 

### Proposed Solution

Add the relevant permissions
If a user has the relevant permissions and they right click another user in a voice channel - they will be given the option to kick them or move them

jagobainda:
Closes #28 
Closes #745 

## Summary
Adds the ability for a moderator to **drag a user from one voice channel to another** in the sidebar. Gated behind a new dedicated `MOVE_MEMBERS` permission.

## How it works
The server stays **client-driven**: `voice.moveUser` only validates and emits a `USER_VOICE_MOVED` event targeted at the moved user; that user's client reuses the existing `joinVoice` + `init` flow to switch channels. No mediasoup runtime is manipulated by hand, and every sidebar updates through the existing `USER_JOIN_VOICE` / `USER_LEAVE_VOICE` events.

## Changes
- **Permission**: new `MOVE_MEMBERS` (`packages/shared`) + translations in the 7 locales. The roles UI picks it up automatically from the enum.
- **Server**: new `voice.moveUser` route + `voice.onMoved` subscription, `USER_VOICE_MOVED` event, own `moveMembers` rate-limiter.
- **Client**: `moveUserToVoiceChannel` action, `useReceiveVoiceMove` hook, and native HTML5 drag-and-drop on voice users / channels.
- **Tests**: guard-path coverage for `voice.moveUser` in `voice.test.ts`.

## Design decisions
- **Dedicated `MOVE_MEMBERS` permission** instead of reusing `MANAGE_USERS`, so "move" can be granted without ban/kick power.
- **Only moving other users** (moderation action); dragging yourself is not in scope.
- **Native HTML5 DnD** isolated to this feature, leaving the existing `@dnd-kit` channel/category reordering untouched (minimal change; works across categories). Unifying both under a single `DndContext` is a possible follow-up.
- **Multi-session safe**: the event carries `fromChannelId`, and only the session actually in the origin channel acts on it.

## Decision to make before merge
Discord, when moving a user between voice channels, checks the **mover's** permissions, a moderator can drop someone into any channel regardless of whether that user could normally join it.

This PR instead respects the **moved user's** permissions: the server validates that the target can actually enter the destination, `VIEW_CHANNEL` + `JOIN` (channel) and `JOIN_VOICE_CHANNELS` (server), mirroring `voice.join`. **If the moved user has no access to that voice channel, they can't be moved there** (the moderator gets an error and the user stays put).

Worth confirming this is the semantics we want before merging.

https://github.com/user-attachments/assets/70d45163-1b21-4f8e-bc96-8397b1092859

## Testing
- Manual E2E: move across categories, move to a restricted channel (rejected), and multi-tab.


pixelsdontmove:
I would argue this function should be more moderator-override than permission-validation.

In the current model, if a moderator wants to bring a guest into a private room to fill up a team in a game, they must go to role settings, find the guests role or create a temporary one, modify permissions to allow VIEW_CHANNEL and JOIN_VOICE for that specific channel. And remove those permissions when no longer needed.

This creates administrative overhead/friction, and could potentially lead to permission creep.

Our community structure relies on a small, permanent core and a larger rotating group of guests.

**The problem:** We want our private channels to remain strictly invisible/unjoinable to guests to maintain privacy and reduce clutter.
**The solution:** The MOVE_MEMBERS permission should act as a temporary bridge. By allowing a moderator to drag a user into a channel they cant normally access, we allow for ad-hoc access that is controlled by human intervention rather than static permission hierarchies.

**Safety:** The concern about security is mitigated by the fact that the action is gated by MOVE_MEMBERS. We are not giving users the ability to bypass permissions, we are giving _trusted moderators_ the ability to easily facilitate communication for that event/game/etc.

**Proposed safety check:** Instead of checking moved _users_ permissions, we should validate that the _moderator_ has the authority to see/join the destination channel.

jagobainda:
@pixelsdontmove This approach seems perfectly reasonable to me, and it's something I had already considered. That's why I added the "Decision to make before merge" section in the description, so that when Diogo reviews it, he can decide which approach he finds more convincing. If needed, I can change it without any problem.

apps/server/src/routers/voice/move.ts
```diff
@@ -0,0 +1,89 @@
...
+        ChannelPermission.VIEW_CHANNEL
+      ),
+      channelUserCan(input.channelId, input.userId, ChannelPermission.JOIN),
+      userCan(input.userId, Permission.JOIN_VOICE_CHANNELS)
+    ]);
+
+    invariant(canView && canJoin && canUseVoice, {
+      code: 'FORBIDDEN',
+      message: 'Target user cannot join the destination channel'
+    });
```
Melonendk: In my opinion.
A moderator who moves should be able to move the user regardless of the user can see the voice channel or not.
My reason is sometimes you will properly have closed spaces and have to invite a person in to it. and if you cant move him up you have to change channel permissions, even if its only a one time thing. which makes it extra steps.

But you laying the groundwork for the permission rework when that happens. So if the user is higher Role (hierarchy) can not be moved. unless has same role. - Depending on what the team wants. - could also be done in server settings. But right now its me rambling :P

apps/server/src/routers/voice/move.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    invariant(currentRuntime, {
+      code: 'BAD_REQUEST',
+      message: 'User is not in a voice channel'
+    });
+
+    invariant(currentRuntime.id !== input.channelId, {
+      code: 'BAD_REQUEST',
+      message: 'User is already in that channel'
+    });
```
Melonendk, line 72: Fine to log as an attempt to move. But otherwise should not do a toast for simple as an attempt and show an error for nothing.

apps/client/src/features/server/voice/actions.ts
```diff
@@ -175,6 +175,21 @@ export const joinVoice = async (
...
+export const moveUserToVoiceChannel = async (
+  userId: number,
+  channelId: number
+): Promise<void> => {
+  const client = getTRPCClient();
+
+  try {
+    await client.voice.moveUser.mutate({ userId, channelId });
+  } catch (error) {
+    toast.error(getTrpcError(error, 'Failed to move user'));
```
Melonendk, line 189: More checks on what type of error. as not all should be applied to a toast. like an simple attempt.

jagobainda:
## Changes after review

I've reworked the move logic following the moderator-override approach proposed by @pixelsdontmove (and seconded by @Melonendk in the review):

- **The target user's channel permissions are no longer checked.** `voice.moveUser` now validates that the **caller** can join the destination channel instead (`needsChannelPermission(JOIN)`, which for private channels also gates on `VIEW_CHANNEL` internally — so the moderator must be able to both see and join it). The moved user only needs the global `JOIN_VOICE_CHANNELS` permission.
- **One-time move grant** (`helpers/voice-move-grants.ts`): when a moderator moves a user, a single-use, channel-specific grant (30s TTL) lets the moved user's client complete the `voice.join` into the destination even if they lack permissions on that channel. Any user-initiated join invalidates the grant, so it can't be hoarded by a modified client to sneak into the private channel later, outside the move flow.
- **Privacy is preserved outside the session**: the hidden channel is published to the moved user while they're connected (`CHANNEL_CREATE`) and unpublished when they leave (`CHANNEL_DELETE`). The sidebar treats the voice channel you're currently connected to as visible even without `VIEW_CHANNEL`.
- **Tests**: updated the existing ones to the new semantics and added coverage for: caller without join permission on the destination, moving a user into a channel they can't access themselves, the one-time grant flow (join fails → move → join succeeds → second join fails), and a regression test pinning the "caller with `JOIN` but without `VIEW_CHANNEL` on a private channel is rejected" guarantee, since that gating is an implicit behavior of `hasChannelPermission` and a future refactor could silently drop it.

## About the error/toast feedback comments

I'm intentionally leaving the toast handling as it is. After this rework, the remaining error paths in `moveUser` (channel not found, not a voice channel, target not in voice, already in that channel, etc.) can't be reached through normal UI interaction, the client already prevents them (drag is disabled without permissions, you can't drop a user into the channel they're already in, and the client aborts stale moves by checking `fromChannelId`). They can only fire when something is genuinely wrong: stale state, a race, or a modified client. In those cases I'd rather surface explicit feedback than fail silently, a toast on a broken flow is useful signal, and suppressing it would just hide the problem. Happy to revisit if it turns out to be noisy in practice.

## New demo


https://github.com/user-attachments/assets/e3174303-0675-49bc-81a6-cc181ba576d9

diogomartino (approved):
Thanks! It needs some changes but I'll make them myself.

jagobainda:
it's nice to see you @diogomartino

PR #764, opened 2026-07-08
fix(#763) update noise suppression to rnnoise

### LINKED ISSUE #763: [Feature]: Noise Suppression should be enabled by default
### Problem

When users join background noise comes through clearly, requiring manual configuration to fix it.

### Proposed Solution

Have noise suppression enabled by default, I have found that RNNoise works best,

Thanks Dev :)

nathangigliotti:
**Summary**
This PR enables RNNOISE noise suppression by default and simplifies the RNNoise audio worklet initialization by always using a 48kHz AudioContext.

**Changes**
1. **Enable RNNOISE as the default noise suppression.**
   - Changed the default device setting for `noiseSuppression` from `NoiseSuppression.NONE` to `NoiseSuppression.RNNOISE`, so new users get noise suppression out of the box.

2. **Simplify & harden RNNoise AudioContext setup.**
   - Removed the `RNNOISE_SAMPLE_RATE` constant and the logic that read the native sample rate from the input device's audio track (`getAudioTracks()[0]?.getSettings().sampleRate`).
   - Removed the runtime error throw when the native sample rate was not 48kHz — this was fragile since device-reported sample rates can vary or be unavailable.
   - The `AudioContext` is now always created with a hardcoded `sampleRate: 48000`, which is what RNNoise requires. This is more reliable and avoids unnecessary failures.

**Why this works:** 
The browser's `createMediaStreamSource` automatically resamples the microphone input to match the context. By forcing a 48kHz context, RNNoise is guaranteed its required sample rate regardless of the microphone's native hardware rate.

Closes #763

### DIFF
--- a/apps/client/src/components/devices-provider/index.tsx
+++ b/apps/client/src/components/devices-provider/index.tsx
@@ -29,7 +29,7 @@ const getDefaultDeviceSettings = (): TDeviceSettings => ({
   webcamResolution: Resolution['720p'],
   webcamFramerate: 30,
   echoCancellation: false,
-  noiseSuppression: NoiseSuppression.NONE,
+  noiseSuppression: NoiseSuppression.RNNOISE,
   autoGainControl: true,
   noiseGateEnabled: false,
   noiseGateThresholdDb: MICROPHONE_GATE_DEFAULT_THRESHOLD_DB,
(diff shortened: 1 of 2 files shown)


jagobainda:
I recommend opening PRs from a branch based on the `development` branch of your fork, for example: `fix/763-update-noise-suppression-to-rnnoise`.

Issue #779, opened 2026-08-09
Audit


diogomartino:
Check the PR for more information: https://github.com/Sharkord/sharkord/pull/780

PR #784, opened 2026-08-19
feat: add server reconnection handling and remove stored credentials

ruedigerlenz:
## Summary

• Automatically retry unexpected server disconnects with backoff
• Show a reconnecting banner while preserving the current app view
• Stop storing user and server passwords locally
• Clean up WebSocket state after failed connections


diogomartino:
Thanks for the PR, but this was recently added in #779.

PR #796, opened 2026-08-29
feat(528): emoji reactions in voice channels

### LINKED ISSUE #528: [Feature]: Emoji reactions in voice channels
### Problem

Participants in a voice channel may often want to express agreement, or disagreement, or surprise, or kudos, etc -- without needing to find a moment to unmute and interject verbally.  A passive communication that doesn't need to interrupt the flow of the conversation in order to be effective.

### Proposed Solution

We could take inspiration from Google Meet's relatively straightforward implementation:

Choosing an emoji reaction shows that emoji on the user's card, which fades away after some seconds.  Optional additions could include playing a relevant sound, and/or having the emoji "flutter" up the screen.

Discord used to have this in a very similar way, but some many months ago they retired it, in favor of the Soundboard feature, which seems overkill, but provides a paid upsell path for them.  

### See also

- Reddit thread where users mourned the loss of this feature in Discord: https://www.reddit.com/r/discordapp/comments/1khph88/emoji_in_voice_channels_is_gone/
- Fake mockup showing "fluttering" up the screen in Google Meet: https://storage.googleapis.com/support-kms-prod/W7EWiptdypbQ2LFR5hn8xIXjp9pe1PlTQBJE

jagobainda:
## Summary

Closes #528

Adds emoji reactions to voice channels, implemented the way the issue proposes: a
reaction picker in the voice controls bar, and the chosen emoji floats up from the
sender's card for everyone in the channel, like the Google Meet screenshot attached to
the issue.

https://github.com/user-attachments/assets/2a086574-7bcd-417f-94d0-49e45e7d670d

**Server.** New `voice.sendReaction` mutation and a `voice.onReaction` subscription.
The mutation reuses `getCurrentVoiceRuntime`, so it only accepts a caller who is
actually connected to a voice channel, validates the emoji against a closed set
(`VOICE_REACTION_EMOJIS` in `packages/shared`) and is rate limited from
`config.rateLimiters.voiceReaction`. Nothing is persisted, the reaction is published
and forgotten.

**Client.** Reactions live in the `server` slice keyed by user, capped per user so a
burst cannot cover a card, and are dropped by a timer once the animation ends. The
store only holds `{ id, emoji }`; the drift, tilt and scale that make a burst look like
separate emoji instead of one column are rolled in the component and kept for the life
of that reaction.

The subscription is a global broadcast rather than channel scoped, on purpose: reactions
render on the cards of whichever voice channel you are looking at, which is not
necessarily the one you are connected to. This matches how `onJoin`, `onLeave` and
`onUpdateState` already work.

## Additional Context

`voice.test.ts` covers the broadcast payload, the rejection of an emoji outside the
allowed set, the rate limiter, and `sendReaction` is added to the shared
`getCurrentVoiceRuntime` guard table so the "not in a voice channel" path is covered
too.


apps/server/src/routers/voice/send-reaction.ts
diogomartino, line 1: Maybe it's a good idea to add a new permission to gate reactions. It can be given by default, but it's a nice to have if admins for some reason don't want this feature or want to gate specific users.

Then you need to add the permission check here in this route.

Something like `Permission.SEND_VOICE_REACTION`

diogomartino (changes requested):
Also, why not allow any emoji as reaction? Even custom ones.

jagobainda:
@diogomartino I based it on how it was described in the issue, the Google Meet screenshot, and I also know that it’s implemented this way in Teams. I didn’t end up doing what you suggested because I thought it might seem like too much, but I’m fine with adding it. Does this approach look good to you?

https://github.com/user-attachments/assets/42707cdd-71d5-410b-ba45-c50cf6d21d1a

diogomartino:
Yea, I never understood why those apps do not let you pick other emojis. IMO that's a pretty stupid limitation and unless I'm missing something really apparent I don't really see why we should limit things to those 4 or 5. It might look a bit strange when an emoji is a rectangle but it's up to the server admins to upload nice emojis I guess.

I'll try to review it this weekend.

PR #802, opened 2026-09-01
chore(786): improve plugin sdk

### LINKED ISSUE #786: [Feature]: Plugin HTTP route authentication
### Problem

Right now there's no way of authenticating a request from inside a plugin route.

### Proposed Solution

Implement that.

diogomartino:
## Summary

Closes #786

This issue was supposed to be only http authentication, but it will be a much bigger thing.

## Checklist

- [x] References an issue that has already been discussed
- [x] Targets the `development` branch
- [x] Covers one feature or fix, not several
- [x] `bun run magic` passes (format, types, lint)
- [x] `bun run test` passes
- [x] Tests added or updated where the change needed them

## Additional Context

This is basically a big improvement on what the plugin system is capable of. The current implementation was basically done based on what I actually needed at the time to make some plugins. This extends that functionality by a lot.

- [x] Audit, refactor and reorganize existing logic
- [x] New hooks: `onBeforeMessageSave`, `onBeforeChannelCreate`, `onBeforeVoiceJoin`, `onBeforeLogin`. Probably gonna add more.
- [x] Special folder to store plugin specific files
- [x] Added `onUpgrade` server hook to allow plugins to make migrations in case they need.
- [x] Data & Write apis extended: messages, roles, users, channels, categories, permissions.
- [x] Added enum type to settings which populates a dropdown
- [x] New permissions system which gates access to actions, commands and ui components.
- [x] Added new UI slots, specially in User Settings. This will allow plugins to manage client side settings more easily.
- [x] HTTP routes authentication: requests can now be authenticated.
- [x] Expose moderation actions (like ban, kick, etc)
- [x] Add (a lot of) missing events
- [x] Server push to client: this will allow plugins to push state directly to a client (not sure if I'll include this one yet)
- [x] Per user plugin storage
- [x] Add more test coverage, specially e2e tests
- [x] Add detailed jsdoc to sdk
- [x] Add support for consumers consumption

Existing plugins will need an update to work with the new sdk version.


EmoticonOrden:
I see a potential in this...

PR #810, opened 2026-09-05
fix: match disconnect button height in voice controls

jagobainda:
## Summary

Sorry for not creating an issue

The disconnect button used `h-full`, which resolves to `auto` when the controls bar is absolutely positioned, so it shrank to its own content height with "always show voice controls" off. Replaced with `h-auto self-stretch` so it matches the controls group in both modes.

### VIDEO: before / after

https://github.com/user-attachments/assets/9be66066-c091-4968-bcdb-df64befc5d46

https://github.com/user-attachments/assets/546d218f-5d06-4c5b-bbbc-2db96a2517e3

### DIFF
--- a/apps/client/src/components/channel-view/voice/controls-bar.tsx
+++ b/apps/client/src/components/channel-view/voice/controls-bar.tsx
@@ -119,7 +119,7 @@ const ControlsBar = memo(({ channelId }: TControlsBarProps) => {
       <Tooltip content="Disconnect">
         <Button
           className={cn(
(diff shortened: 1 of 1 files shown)


Melonendk:
Haha welp, didn see that tho but lucky its an easy fix. - Recommendation, check if its fixed in multiple browsers. - chrome mozilla edge.
But nice!

jagobainda:
Checked in Chrome, Firefox, and Brave!

diogomartino (approved):
Ty!

Issue #813, opened 2026-09-12
[Bug]: Overriding DATA_DIR via env variable causes permission and directory layout issues


kaiyga:
### Before reporting

- [x] I searched the existing issues and this has not been reported yet
- [x] I can reproduce this on the latest release

### Sharkord Version

0.0.25

### How are you running Sharkord?

Docker

### Environment

Describe the bug
Currently, docker-entrypoint.sh allows overriding DATA_DIR via environment variables (e.g., DATA_DIR=${DATA_DIR:-"/home/bun/.config/sharkord"}). However, overriding this path breaks container execution because:

Setting DATA_DIR to a custom location leaves /home/bun/.config without proper ownership (chown), causing Permission denied errors for the bun user when the application expects system configuration in .config.

The application expects configuration files strictly inside /home/bun/.config, making DATA_DIR overrides unstable and unsupported.

Expected behavior
Since the container relies on a fixed directory structure, DATA_DIR should be hardcoded to /home/bun/.config/sharkord, and ownership should be recursively applied to /home/bun/.config to ensure stable operation.

Proposed Fix
Hardcode DATA_DIR in docker-entrypoint.sh and update chown to target /home/bun/.config.

### Steps to Reproduce

Set DATA_DIR and see that nothing changed;D

### Expected Behavior

I exept that i can set a data_dir point in my container

### Actual Behavior

Ignore DATA_DIR 

from -e and envs in docker-compose

### Logs

```shell

```

PR #814, opened 2026-09-12
[FIX]: #813 Set changable DATA_DIR variable for flexible configuration of docker

kaiyga:
Description: Issue #813

Summary of changes
This PR fixes runtime permission errors and inconsistent directory behavior in the Docker container by enforcing a fixed data path and correcting ownership 

What was done:

Hardcoded DATA_DIR: Removed the ability to override DATA_DIR via environment variables. Allowing custom paths broke application execution because the runtime relies on hardcoded paths inside /home/bun/.config/sharkord.

Fixed parent directory permissions: Updated chown -R to target /home/bun/.config instead of just $DATA_DIR. This ensures that the non-root bun user has full write permissions to the entire configuration root directory, preventing Permission denied errors.

How to test:

Build and run the Docker image.

Verify that the container initializes properly under the bun user without permission errors in /home/bun/.config.


docker-entrypoint.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -e
 
-DATA_DIR="/home/bun/.config/sharkord"
+DATA_DIR=${DATA_DIR:-"/home/bun/.config/sharkord"}
```
1816586742-stack, line 4: The name this reads is not the one the server reads, so a container that sets it gets the failure the PR is trying to fix.

```diff
-DATA_DIR="/home/bun/.config/sharkord"
+DATA_DIR=${DATA_DIR:-"/home/bun/.config/sharkord"}
```

The server looks for `SHARKORD_DATA_PATH` and never `DATA_DIR`:

```ts
// apps/server/src/helpers/paths.ts:15
const INJECTED_DATA_PATH = process.env.SHARKORD_DATA_PATH;

if (INJECTED_DATA_PATH) return path.resolve(INJECTED_DATA_PATH);
// ...
return path.join(getAppDataPath(), 'sharkord');   // $XDG_CONFIG_HOME or ~/.config
```

```ts
// apps/server/src/declarations.d.ts:36
SHARKORD_DATA_PATH?: string;
```

```console
$ git grep -n 'env.*DATA_DIR' -- src-tauri apps/server
(no matches)
```

So with `DATA_DIR=/data` the entrypoint creates and chowns `/data`, while the app still resolves `/home/bun/.config/sharkord` and writes there. The one directory the app actually uses is now the one nothing chowns, because the second hunk below moves the `chown` off `/home/bun/.config` and onto `$DATA_DIR` - which is the opposite of what the report describes as broken.

That second hunk is the reason this is worth catching before merge rather than after:

```diff
 mkdir -p "$DATA_DIR"
-chown -R bun:bun /home/bun/.config
+chown -R bun:bun "$DATA_DIR"
```

`/home/bun/.config` is the parent the app reads and writes `config.ini` beside, so dropping it from the chown removes an ownership pass that the default path depends on.

## Relationship to #823

The two PRs are in the same lane and take opposite approaches, so flagging that rather than leaving it to be discovered: #823 resolves `SHARKORD_DATA_PATH` in the entrypoint (the name the app reads) and leaves the default `chown` of `/home/bun/.config` in place, on the reading that #813 asks for the variable to work - the report says "I exept that i can set a data_dir point in my container" - rather than to be removed.

Only one of the two should land. If the intended answer is the hardcoded path, then the entrypoint cannot stay as it is either: with the current literal, `DATA_DIR` does nothing, and the app already honours `SHARKORD_DATA_PATH`, so the two halves disagree until one of them is changed.

